PR #3, opened 2019-07-16
Add support for cargo doc

basvandijk:


build.nix
```diff
@@ -3,7 +3,10 @@ src:
   cargoBuild
 , #| What command to run during the test phase
   cargoTest ? "cargo test --$CARGO_BUILD_PROFILE"
+, #| What command to run during the optional doc phase
+  cargoDoc ? "cargo doc --offline --frozen"
```
nmattia: `--frozen` _may_ fail in some circumstances when cargo needs to update the Cargo.lock for weird reasons, even though it won't access the network

build.nix
```diff
@@ -149,6 +155,17 @@ with rec
             runHook postCheck
           '';
 
+
+        docPhase = lib.optionalString doDoc ''
+          runHook preDoc
+
+          echo "Running doc command:"
+          echo "  ${cargoDoc}"
+          CARGO_TARGET_DIR=target_doc ${cargoDoc}
```
nmattia: Do you know what the doc layout looks like in the `target` dir? Maybe we should just share the same `CARGO_TARGET_DIR` as the rest of the build (to avoid weird rebuilds) and simply copy out the doc-related files

build.nix
```diff
@@ -183,6 +200,11 @@ with rec
             mkdir -p $out
             cp -r target $out
 
+            ${lib.optionalString doDoc ''
+            mkdir -p $doc
+            cp -r target_doc/doc/* $doc
```
nmattia: ah, looks like all the doc ends up in `target/doc`. Might as well use a single target dir then. What do you think?

basvandijk: I used a dir separate from `target` since the line `cp -r target $out` above would copy the whole `target` to `$out` and I would like to have the docs only the separate output `$doc`. 

Maybe we can change `cp -r target $out` to only copy the stuff that you actually need in `$out`.

PR #10, opened 2019-08-23
Remove references to the source code from the generated cargo docs

basvandijk:


build.nix
```diff
@@ -176,6 +194,12 @@ with rec
           echo "  ${cargoDoc}"
           ${cargoDoc}
 
+          ${lib.optionalString removeReferencesToSrcFromDocs ''
+          # Remove references to the source derivation to reduce closure size
+          find target/doc -name "*\.rs\.html" -exec \
+            perl -pi -e 's/<meta name="description" content="Source to the Rust file .*?">/<meta name="description" content="Source to the Rust file removed to reduce Nix closure size.">/' {} +
```
basvandijk: I need `perl` instead of `sed` for non greedy matching. See: https://stackoverflow.com/a/1103177

basvandijk:
@nmattia I removed the dependency on Perl.

PR #14, opened 2019-09-22
format with nixpkgs-fmt

zimbatm:
What do you think of the format @nmattia ?

### DIFF
--- a/builtins/default.nix
+++ b/builtins/default.nix
@@ -7,54 +7,57 @@
 
 rec
 {
-    toTOML = import ./to-toml.nix { inherit lib; };
-    writeTOML = name: attrs: writeText name (toTOML attrs);
+  toTOML = import ./to-toml.nix { inherit lib; };
+  writeTOML = name: attrs: writeText name (toTOML attrs);
 
-    readTOML = usePure: f:
-      if usePure then
-        builtins.fromTOML (builtins.readFile f)
-      else
-        builtins.fromJSON (builtins.readFile (
+  readTOML = usePure: f:
+    if usePure then
+      builtins.fromTOML (builtins.readFile f)
+    else
+      builtins.fromJSON (
+        builtins.readFile (
           runCommand "from-toml"
-          { buildInputs = [ remarshal ];
-            allowSubstitutes = false;
-            preferLocalBuild = true;
-          }
-          ''
-            echo "$from_toml_in" > in.toml
-            remarshal \
-              -if toml \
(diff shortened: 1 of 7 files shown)


nmattia:
I like it! The latest changes actually are in https://github.com/nmattia/naersk/tree/nm-cargo-release, let me merge that for you first.

nmattia:
https://github.com/nmattia/naersk/pull/16

builtins/default.nix
```diff
@@ -7,54 +7,57 @@
...
+  pathExists = path:
+    let
+      all = lib.all (x: x);
+      isOk = part:
+        let
+          dir = builtins.dirOf part;
+          basename = builtins.unsafeDiscardStringContext (builtins.baseNameOf part);
+          dirContent = builtins.readDir dir;
+        in
+          builtins.hasAttr basename dirContent && # XXX: this may not work if the directory is a symlink
```
nmattia: @zimbatm looks like this guy got moved! Issue when formatting infix operators?

builtins/to-toml.nix
```diff
@@ -25,106 +25,115 @@ let
...
-    if typeOf x == "set" then
-      if lib.isDerivation x then "string" else "set" else
-    if typeOf x == "list" then
-      if length x == 0 then "list"
-      else
-        let ty = typeOf (elemAt x 0); in
-        #assert (all (v: typeOf v == ty) x);
-        if ty == "set" then "list_of_attrs" else "list"
-    else abort "Not implemented: toml type for ${typeOf x}";
+      if typeOf x == "bool" then "bool" else
```
nmattia: mh this is a bit annoying, it gets much harder to read

nmattia: according to some of the formatting later on in this module, putting the `else` on the next line should fix it

builtins/default.nix
zimbatm: hmm maybe. I'll just fix it manually for now

builtins/to-toml.nix
zimbatm: what do you think of the new format?

builtins/default.nix
zimbatm: moved the code around a bit

zimbatm:
@nmattia pushed an update with the latest greatest version of nixpkgs-fmt. I skipped nix/sources.nix

nmattia:
Looks great, thanks @zimbatm !

PR #15, opened 2019-09-22
Src attr

zimbatm:
What do you think of this @nmattia ?

Based on top of #14 

This changes how the src argument is taken so it's part of the attrs. The main advantage is that nix will fail saying that `src` is required. It also makes it a bit closer to how it's done in nixpkgs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,13 @@ And then
 
 ``` nix
 let
-    pkgs = import <nixpkgs> {};
-    sources = import ./nix/sources.nix;
-    naersk = pkgs.callPackage sources.naersk {};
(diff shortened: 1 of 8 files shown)


nmattia:
>  It also makes it a bit closer to how it's done in nixpkgs.

I know, and I don't like it :) there'll always be a `src`, I don't want users to have to think about anything. If you _really_ want `src` then I'd suggest having `buildPackage` take either a path/drv/etc _or_ an attrset, but that can be a bit confusing.

zimbatm:
okay :) I think you'll change your mind later but I won't insist :p

PR #16, opened 2019-09-22
Bunch of improvements

nmattia:
* fix tests on Linux
* improve the per-crate build to not recompile on install and ...
* ... actually drop the per-crate build, thanks @basvandijk 
* run clippy and other on deps (which was then removed)
* fix `--release` flag

Let's see what we broke!

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -224,7 +224,6 @@ with rec
             done
 
             mkdir -p $out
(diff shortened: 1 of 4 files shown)


nmattia:
Got a few leftover foos and bars but this is clearly an improvement. Let's see what happens.

PR #17, opened 2019-10-03
cargo build/test with --frozen to ensure Cargo.lock is up to date

basvandijk:
For example if I change the name `simple-dep` to `simple_dep` in
`test/simple-dep/Cargo.toml` and build without this commit I don't get
any errors. If I run with this commit I get:

```
cargo build --release --frozen -j 6 -Z unstable-options --out-dir out
error: the lock file /private/var/folders/5h/b11m6fxj2tqgbxwz5bgjy42c0000gn/T/nix-build-some-name.drv-0/dummy-src/Cargo.lock needs to be updated but --frozen was passed to prevent this
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -96,10 +96,14 @@ with rec
             find . -type f -name '*.rs' -exec touch {} +
           '';
           cargoBuild = attrs.cargoBuild or ''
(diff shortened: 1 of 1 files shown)


nmattia:
@basvandijk closing for now, feel free to reopen

PR #26, opened 2019-10-10
Simplify configuration for single step build

nmattia:
There's now a single function `buildPackage` that one can call with `singleStep = true` to skip the pre build ("some-name").

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -45,6 +45,9 @@ with rec
         { usePureFromTOML = attrs.usePureFromTOML or true;
           readTOML = builtinz.readTOML usePureFromTOML;
 
+          # Whether we skip pre-building the deps
+          isSingleStep = attrs.singleStep or false;
+
           # The members we want to build
           wantedMembers =
             lib.mapAttrsToList (member: _cargotoml: member) wantedMemberCargotomls;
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -148,10 +141,10 @@ with rec
...
                   )
-                )];
+                );
             });
   };
 
-{ inherit buildPackageSingleStep buildPackageIncremental crates;
-  buildPackage = buildPackageIncremental;
+{
+  inherit buildPackage;
```
basvandijk: I wondering, since `naersk` now exports just this single `buildPackage` function, have you considered just returning `buildPackage` instead of `{ inherit buildPackage; }`?

nmattia: So that'd look like `let pkg = (import naersk {}) src attrs; in ...` ?

basvandijk: Yes and you would typically assign the function to a name like:

```
let 
  naerskBuildPackage = import naersk {};
  pkgs = naerskBuildPackage src attrs;
in ...
```

Although that's  not that different than the current:

```
let 
  naersk = import naersk {};
  pkgs = naersk.buildPackage src attrs;
in ...
```

However, if users want that they can always introduce the name spacing themselves like:

```
let 
  naersk = { buildPackage = import naersk {}; };
  pkgs = naersk.buildPackage src attrs;
in ...
```

Oh well, I'll let you decide ;-)

nmattia: Good point. I'll avoid breaking compatibility for now but it'd be good to have a think at some point.

PR #31, opened 2019-10-29
Copy paths from the [patch] section into the dummy-src

basvandijk:
This allows users to override the source of crates using:

```
[patch.crates-io]
bar = { path = 'my/local/bar' }
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -80,6 +80,17 @@ with rec
               )
             );
 
+          patchedSources =
+            let
+              mkRelative = po:
+                if lib.hasPrefix "/" po.path
+                then throw "'${toString src}/Cargo.toml' contains the abolsute path '${toString po.path}' which is not allowed under a [patch] section by naersk. Please make it relative to '${toString src}'"
+                else src + "/" + po.path;
+            in lib.optionals (builtins.hasAttr "patch" toplevelCargotoml)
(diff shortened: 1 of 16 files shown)


default.nix
```diff
@@ -80,6 +80,12 @@ with rec
               )
             );
 
+          patchedSources =
+            lib.optionals (builtins.hasAttr "patch" toplevelCargotoml)
+              (map (po: src + "/" + po.path)
+                (lib.collect (as: lib.isAttrs as && builtins.hasAttr "path" as)
```
nmattia: Do you know what else than `crates.io` can be patched?  Is it safe to just assume that we'll "vendor" all patches that have a `path`?

basvandijk: I think it's safe.

default.nix
```diff
@@ -80,6 +80,12 @@ with rec
               )
             );
 
+          patchedSources =
+            lib.optionals (builtins.hasAttr "patch" toplevelCargotoml)
+              (map (po: src + "/" + po.path)
```
nmattia: note: would be nice to crash if the path is absolute

nmattia:
Does this work with non-virtual manifests? can you add some tests ? (ideally virtual + non-virtual)

PR #33, opened 2019-10-30
Clean up, take 2

nmattia:
* general build code cleanup
* reduce number of derivations built
* remove occurrences of `foo`, `bar` and other silly names
* set name and version based on package metadata

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -7,8 +7,10 @@ with
   };
 rec
 {
-    # creates an attrset from package name to package version + sha256
-    # (note: this includes the package's dependencies)
+    # The list of _all_ crates (incl. transitive dependencies) with name,
+    # version and sha256 of the crate
+    # Example:
+    #   [ { name = "wabt", version = "2.0.6", sha256 = "..." } ]
     mkVersions = cargolock:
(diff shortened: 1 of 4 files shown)


lib.nix
```diff
@@ -99,14 +101,17 @@ rec
           in attrs // lib.optionalAttrs (lib.hasAttr "package" attrs) {
                         package = removeAttrs attrs.package [ "build" ];
                       };
-        cargotomlss = writeText "foo"
-          (lib.concatStrings (lib.mapAttrsToList
-            (k: v: "${k}\n${builtinz.writeTOML "Cargo.toml-ds-aa" (fixupCargoToml v)}\n")
+
+        # a list of tuples from member to cargo toml:
+        #   "foo-member:/path/to/toml bar:/path/to/other-toml"
+        cargotomlss = lib.concatStringsSep " " (lib.mapAttrsToList
```
basvandijk: Note that [Nix will automatically intersperse with " " when you pass a list of things to a derivation](https://nixos.org/nix/manual/#ssec-derivation) so we could simplify this to:
```suggestion
        cargotomlss = (lib.mapAttrsToList
```

PR #34, opened 2019-11-01
Remove `name` from `attrs` for the deps build

basvandijk:
Otherwise the nicely generated:

`pname = "${packageName}-deps"`

gets overridden by `name` if the user specifies a name.


nmattia:
then maybe we should allow the user to specify either `name` or `pname`?

nmattia:
Will need to be rebased. Closing for now, as it's unclear which `name` we want to appear in the derivation.

PR #46, opened 2019-12-13
Add support for git dependencies.

nmattia:
**NOTE: implementation was changed drastically, see comments below**

Because this is highly experimental, this needs to be enabled in the config with `allowGitDependencies = true`.

The relevant commit for git dependencies: https://github.com/nmattia/naersk/commit/ea1555099f9f4a9f75309d11ec12e4b455a054c5

The other commits are preparation cleanup.

This works by first fetching the git dependencies with `fetchGit`, and then patching the `Cargo.toml`. A few notes:

* the .cargo/config `paths` does not work, because cargo doesn't register it until _after_ it has tried (and failed) to checkout the deps from the server
* the `replace` mechanism of `Cargo.toml`s doesn't work, same reason as above
* mimicking the `$CARGO_HOME/git/{db,checkout}` _seems_ fairly simple (the format is `{db,checkout}/<crate>-<siphash_of_url>/...` but it turns out cargo (libgit2 really) expects a full checkout.

CC @basvandijk @Profpatsch @curiousleo


nmattia:
Found a much simpler, much more robust way of implementing this.

nmattia:
@basvandijk occurred to me that I could just copy what `cargo vendor` does; not sure why I didn't think about it the other day.

Profpatsch:
Can we upstream some of this so it reduces complexity for us in the future?

nmattia:
@profpatsch upstream where? Naersk or the git deps?

Profpatsch:
Upstream to cargo.

PR #51, opened 2020-01-10
Fix failure on duplicate git dependencies

nmattia:
The git dependencies were indexed by crate name, which is an issue if a dependency is imported through git at various commits. Moreover, if a crate is imported several time at _the same commit_, it is now unpacked only once.


build.nix
```diff
@@ -89,17 +89,25 @@ let
         checkout=$(echo "$dep" | jq -cMr '.checkout')
         url=$(echo "$dep" | jq -cMr '.url')
         tomls=$(find $checkout -name Cargo.toml)
+        rev=$(echo "$dep" | jq -cMr '.rev')
+        echo "DEP IS $dep"
```
knl: Maybe to help future debugging?
```suggestion
        echo "Dependency: $dep (rev: $rev from $url) checked out at $checkout"
```

build.nix
```diff
@@ -89,17 +89,25 @@ let
...
             | sed -n -e '/\[package\]/,$p' \
             | grep -m 1 "^name\W" \
             | grep -oP '(?<=").+(?=")' \
             || true)
           if [ -n "$name" ]; then
-            echo "$url Found crate '$name'"
-            cp -r $(dirname $toml) $out/$name
-            chmod +w $out/$name
-            echo '{"package":null,"files":{}}' > $out/$name/.cargo-checksum.json
+            key="$name-$rev"
```
knl: Do you maybe want to have a version in here as well? I see that people write in their `Cargo.toml` stuff like:

    "prometheus 0.8.0 (git+https://github.com/tikv/rust-prometheus?rev=d919ccd35976b9b84b8d03c07138c1cc05a36087)"

Where that `0.8.0` is totally arbitrary version. And it might happen that several projects use different numbers.

build.nix
nmattia: wops! left over. thanks!

build.nix
nmattia: This is actually a bit tricky; getting the name already involves some `grep` black magic because we need to grab it from the `toml` file; we'd need to do the same for the `version`. Moreover the `key` is only meant to identify a particular version of a crate; `name` + `revision` should be enough.

PR #52, opened 2020-01-10
Exclude .so files from exe copy

pikajude:


nmattia:
@pikajude do you know in what situations cargo generates `so`s? I thought it mostly produced `rlib`s. Also, should we also filter out `.dylib`s?

pikajude:
They're generated when a crate in the workspace exports a proc macro. Is naersk designed to be a first-class Rust crate builder or is it just meant to be a top-level wrapper to produce binaries/libs from a Rust project? If 1), we should keep the shared objects since they would break exported proc macros. If 2), they should all be deleted.

nmattia:
Thanks!

PR #55, opened 2020-01-15
Basic flake.nix

jD91mZM2:
This allows you to use naersk more easily via the current version of [nix-flakes](https://youtu.be/UeBX7Ide5a0) (though it may rapidly change syntax, maybe it should be on its own branch for now) :)  
Btw, I saw your speech on NixCon via YouTube, it's really good!

Before:

```nix
{
  inputs.naersk = {
    url   = "github:nmattia/naersk";
    flake = false;
  };

  # Further down, inside outputs:
  packages = forAllSystems (system: {
    rnix-lsp = let
      naerskBuilt  = nixpkgs.legacyPackages."${system}".callPackage naersk {};
    in naerskBuilt.buildPackage {
      name = "rnix-lsp";
      src  = cleanSourceSomehow ./.;
      root = ./.;
    };
  });
}
```

After:

```nix
{
  inputs.naersk = {
    url   = "github:nmattia/naersk";
  };

  # Further down, inside outputs:
  packages = forAllSystems (system: {
    rnix-lsp = naersk.lib."${system}".buildPackage {
      name = "rnix-lsp";
      src  = cleanSourceSomehow ./.;
      root = ./.;
    };
  });
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docparse/target
+/flake.lock
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,12 @@
+{
+  description = "Build rust crates in Nix. No configuration, no code generation. IFD and sandbox friendly.";
+
+  edition = 201909;
+
+  outputs = { self, nixpkgs }: let
+    forAllSystems = nixpkgs.lib.genAttrs [ "x86_64-linux" "x86_64-darwin" "i686-linux" "aarch64-linux" ];
+  in {
+    # Naersk is not a package, not an app, not a module... It's just a Library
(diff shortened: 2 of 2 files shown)


jD91mZM2:
What do you mean? [Here's a flake that uses the patched naersk](https://github.com/nix-community/rnix-lsp/blob/0a5c0a7c1dd1f47c40367573b94264d912da59d2/flake.nix), and you can use it using
```
nix app github:nix-community/rnix-lsp
```
which automagically runs the rnix-lsp executable thanks to the app section. You can also install the package itself using `nix profile install github:nix-community/rnix-lsp`.

I hope that answered your question?

nmattia:
I don't understand everything but... looks legit! thanks!

PR #60, opened 2020-02-06
Refactor config

nmattia:
TODO:
- [x] update documentation


docparse/src/main.rs
```diff
@@ -42,34 +42,48 @@ fn print_mk_attrs(mk_attrs: SyntaxNode) {
...
-            .expect("Is not OrDefault")
-            .default()
-            .and_then(|def| {
-                let shown = format!("{}", def);
-                if shown != "null" {
-                    Some(shown)
+        let v = e.value();
+        let mextra_note = if let Some(x) = v.clone().and_then(OrDefault::cast) {
+            x.default().and_then(|def| {
+                //let extra_note = format!("Default: `{}`", def);
```
nmattia: remove

nmattia:
That's a very good point. Will need to think about it. If we move to only functions, then I'll still keep the function-or-string logic and throw a meaningful error in case the arg is not a function.

PR #67, opened 2020-02-11
Add GitHub Actions linux check

nmattia:
Add actual bashlib code

Skip bashlib and move straight to install-nix

Install nix in install-nix

Fix install-nix broken string

Run tests on GitHub Actions

Simplify install-nix

Try cachix setup

Parallelize cache upload

Actually run tests

Remove parallel workaround


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+  push:
+    branches:
+      - '*'
+
+jobs:
+  install-nix:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: './.github/actions/nix'
```
zimbatm:
```suggestion
      - uses: cachix/install-nix-action@v6
```
If you don't want to bother doing it yourself. See https://github.com/cachix/install-nix-action

There is also https://github.com/cachix/cachix-action that might be useful.

nmattia: see https://github.com/nmattia/naersk/pull/67/files#diff-eff5c19ca00944a08dae2516e088448bR3, I'm not using it on purpose because I don't want all those node_modules

PR #74, opened 2020-02-18
Remap source path prefix

### LINKED ISSUE #71: Compiled binaries depend on the crate source code
When I build my [rust project](https://github.com/Xe/withinbot) as a Docker image, I get a ~200 MB image. This seems to be because every single one of the crates it depends on gets included into the resulting binary. Here is how I am building a docker image:

- https://github.com/Xe/withinbot/blob/master/docker.nix
- https://github.com/Xe/withinbot/blob/master/default.nix

How do I remove references to the crates from the compiled binary? The paths to the crates seem to live in the Rust panicinfo crud.

andrew-d:
This ensures that the final build artifact(s) don't depend on the entire original source code via paths in panic information.

Fixes #71 

cc @Xe

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -121,6 +121,12 @@ let
     } | jq -cMr '.[]')
     '';
 
+  nixSourcesDir = symlinkJoin {
+    name = "crates-io";
+    paths = map (v: unpackCrate v.name v.version v.sha256)
+      crateDependencies ++ [ unpackedGitDependencies ];
+  };
+
   drv = stdenv.mkDerivation {
     name = "${pname}-${version}";
     inherit
@@ -136,11 +142,7 @@ let
       source = {
         crates-io = { replace-with = "nix-sources"; };
         nix-sources = {
-          directory = symlinkJoin {
-            name = "crates-io";
-            paths = map (v: unpackCrate v.name v.version v.sha256)
-              crateDependencies ++ [ unpackedGitDependencies ];
-          };
+          directory = nixSourcesDir;
         };
       } // lib.listToAttrs (
         map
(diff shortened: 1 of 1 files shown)


Xe:
This fixes it perfectly for me. Thanks!

nmattia:
@andrew-d thanks for the fix! Setting `RUSTFLAGS` will break a few builds at $WORK, I wonder if there's another way to set this up.

Xe:
Maybe have a "removeCrateReferences" boolean like  Go's package support has a "removeGoReferences" boolean?

nmattia:
Ok, thought about this a bit more. here's a plan:

* Instead of using `RUSTFLAGS`, we can use `CARGO_BUILD_RUSTFLAGS`. I use `CARGO_BUILD_RUSTFLAGS` in other projects because _I believe_ it takes precedence over `RUSTFLAGS`. I can't remember the exact reason, but I _think_ it's saner than `RUSTFLAGS`.
* I'll add `CARGO_BUILD_RUSTFLAGS` to the config; at the same time I'll change the `buildPackage` interface to allow passing in any env variables (much like `mkDerivation`). Right now the only way to set an environment variable is through the `override` option, which seemed like a good idea, but in practice is terrible.
* `CARGO_BUILD_RUSTFLAGS` will accept both a function and a list. When it's a list, it's appended to the default options. When it's a function, it takes a default value and generates a new list. The default value will be the above by default, or will be `[]` when (the new flag) `removeCrateReferences` is set to false.

@andrew-d thoughts? I'll prepare the change, then cherry-pick your commit. I'll also re-open a PR because I have no idea how to have GitHub Actions run of PRs from forks.

CC @basvandijk

nmattia:
Here's the first step, forwarding the user attributes: https://github.com/nmattia/naersk/pull/75

nmattia:
Closing in favor of https://github.com/nmattia/naersk/pull/76

I decided to do this dynamically, depending on which environment variables are already set. One of RUSTFLAGS or CARGO_BUILD_RUSTFLAGS is ignored when the other one is set, so by using whichever one is set, we cover most cases.

nmattia:
(Had to re-open a PR for the tests to run; I rebased and kept your commit @andrew-d )

andrew-d:
Sounds good to me - I really like how clean #75 is, and the fix in #76 looks good as well. Thanks for the quick response here 😀

PR #76, opened 2020-02-18
Remap source path prefix

### LINKED ISSUE #71: Compiled binaries depend on the crate source code

nmattia:
This adds the option `remapPathPrefix` which, when set to `true`, reduces the closure size by avoiding references to the source files in the nix store.

Closes #71 

Was #74 

CC @andrew-d

try this out: `niv update naersk -b nm-strip-refs`


build.nix
```diff
@@ -220,6 +226,25 @@ let
...
+      # Remove the source path(s) in Rust
+      if [ -n "$RUSTFLAGS" ]; then
+        RUSTFLAGS="$RUSTFLAGS --remap-path-prefix $crate_sources=/sources"
+        log "RUSTFLAGS (updated): $RUSTFLAGS"
+      elif [ -n "$CARGO_BUILD_RUSTFLAGS" ]; then
+        CARGO_BUILD_RUSTFLAGS="$CARGO_BUILD_RUSTFLAGS --remap-path-prefix $crate_sources=/sources"
+        log "CARGO_BUILD_RUSTFLAGS (updated): $CARGO_BUILD_RUSTFLAGS"
+      else
+        export CARGO_BUILD_RUSTFLAGS="--remap-path-prefix $crate_sources=/sources"
+        log "CARGO_BUILD_RUSTFLAGS (updated): $CARGO_BUILD_RUSTFLAGS"
```
basvandijk:
```suggestion
        log "CARGO_BUILD_RUSTFLAGS: $CARGO_BUILD_RUSTFLAGS"
```

build.nix
```diff
@@ -220,6 +226,25 @@ let
       log "cargo_test_options: $cargo_test_options"
       log "cargo_bins_jq_filter: $cargo_bins_jq_filter"
       log "cargo_build_output_json (created): $cargo_build_output_json"
+      log "crate_sources: $crate_sources"
+      log "RUSTFLAGS: $RUSTFLAGS"
+      log "CARGO_BUILD_RUSTFLAGS: $CARGO_BUILD_RUSTFLAGS"
+
+      ${lib.optionalString remapPathPrefix ''
+
+      # Remove the source path(s) in Rust
```
basvandijk: Maybe indent this block so that you can clearly see where the block begins and ends. But it's style so your call.

nmattia: let's ask `nixpkgs-fmt`

build.nix
nmattia: what do you mean? I'm printing a non-updated version above, here I want to point out that it's been updated.

build.nix
nmattia: `nixpkgs-fmt` is happy, though it does complain about `symlinkJoinPassViaFile` _cough cough_

PR #82, opened 2020-03-06
Set the cargo test option --test-threads to $NIX_BUILD_CORES

basvandijk:
Because it defaults to the number of cores which is often too much parallelism.


basvandijk:
This is causing problem in cargo tests which don't accept the `--test-threads` argument.

Closing in favor of: https://github.com/nmattia/naersk/pull/83.

PR #84, opened 2020-03-13
Fetch git specify ref

basvandijk:
This is to work around a bug in `builtins.fetchGit` where
specifying a `rev` which is not an ancestor of HEAD
results in a git error. When setting `ref` to `rev` we
force Nix to fetch the `rev`.

TODO: This hack can be removed once the following is
merged and released: NixOS/nix#3408

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -218,12 +218,18 @@ let
         cargo_message_format="json"
       fi
 
+      # Rust's `libtest` defaults to running tests in parallel and uses as many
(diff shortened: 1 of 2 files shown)


basvandijk:
This was more of an experiment and I intend to remove this.

I think the longer term solution is to support `branch` and `tag` attributes in git sources like I implemented here: https://github.com/nmattia/naersk/pull/85

nmattia:
Gotcha, closing this then!

Issue #53, opened 2020-01-11
Cannot detect Cargo.toml's [workspace] with nested members declaration


Rizary:
Hi, I am using niv to add https://github.com/rust-analyzer/rust-analyzer/ into my project.

Then, I have the following code:
```
rust-analyzer-src = (import ./nix/sources.nix).rust-analyzer;
rust-analyzer = pkgs.naerks.buildPackage rust-analyzer-src;
```
into my `shell.nix`. However, when I try to run `nix-shell` I got the following error:

```opening file ' nix/store/r193....-source/crates/*/Cargo.toml': No such file or directory``` while the `crates` folder exists in the `rust-analyzer` package.

**Comment by Rizary** [comment]
For more information, here is the screenshot of the error:

![image](https://user-images.githubusercontent.com/7221768/72196060-40f04800-3448-11ea-825d-7fd249ea72be.png)

**Comment by Rizary** [comment]
Digging further, I think the problem lies in this code:

https://github.com/nmattia/naersk/blob/master/config.nix#L173-L178 of

```
    cargotomls =
      let
        readTOML = builtinz.readTOML usePureFromTOML;
      in
        { "." = toplevelCargotoml; } // lib.optionalAttrs isWorkspace
          (
            lib.listToAttrs
              (
                map
                  (
                    member:
                      {
                        name = member;
                        value = readTOML (root + "/${member}/Cargo.toml");
                      }
                  )
```

And in the rust-analyzer's Cargo.toml, the workspace contain `"crates/*"` value:
```
  
[workspace]
members = [ "crates/*", "xtask/" ]
```

Which reads literally `"crates/*"` by this `readTOML (root + "/${member}/Cargo.toml");`

Rizary:
### [Update]

I try to create some workaround like the following:

![image](https://user-images.githubusercontent.com/7221768/72199860-c4279300-3474-11ea-8e64-9069cd56d0bd.png)


But then I got the Error:

`Error: evaluation aborted with the following error message: 'Not implemented: type int'`

And I have no idea where to look at it in the code/rust-analyzer.

nmattia:
> the workspace contain "crates/*" value:

I had no idea that was possible! I'll fix this on Monday.

[...]

PR #86, opened 2020-03-18
Make sure `expandMember` works with a trailing slash

sean-bennett112:
I ran across this issue when trying to build [nushell](https://github.com/nushell/nushell/tree/1ec2ec72b5d772db8759905e03cb34b6409f5067). Their [Cargo.toml](https://github.com/nushell/nushell/blob/1ec2ec72b5d772db8759905e03cb34b6409f5067/Cargo.toml#L16) has a trailing slash in the globbing pattern, which breaks naersk with the same error seen in #53 

This fix addresses that specific issue, however the `members` field actually accepts a much more extensive [globbing syntax](https://docs.rs/glob/0.3.0/glob/struct.Pattern.html), at least according to the docs.

I'm hoping that more involved cases are relatively rare in practice. If anyone has a more extensive/less error-prone way of handling this, please do!

### DIFF
--- a/.github/actions/nix/install-nix
+++ b/.github/actions/nix/install-nix
@@ -62,7 +62,7 @@ save_cache() {
     echo "Using cache '$cachix_cache' for '$GITHUB_REPOSITORY'"
 
     set +x
-    if [ -z "${CACHIX_SIGNING_KEY:-}" ] && [ -n "${INPUT_CACHIX_SIGNING_KEY:-}" ]; then
+    if [ -z "${CACHIX_SIGNING_KEY:=}" ] && [ -n "${INPUT_CACHIX_SIGNING_KEY:-}" ]; then
         echo "CACHIX_SIGNING_KEY not set, but INPUT_CACHIX_SIGNING_KEY is present"
         echo "setting CACHIX_SIGNING_KEY"
         export CACHIX_SIGNING_KEY="$INPUT_CACHIX_SIGNING_KEY"
--- a/config.nix
+++ b/config.nix
@@ -258,7 +258,7 @@ let
         # this turns members like "foo/*" into [ "foo/bar" "foo/baz" ]
         # as in https://github.com/rust-analyzer/rust-analyzer/blob/b2ed130ffd9c79de26249a1dfb2a8312d6af12b3/Cargo.toml#L2
         expandMember = member:
-          if lib.hasInfix "*" member
+          if (lib.hasSuffix "/*" member) || (lib.hasSuffix "/*/" member)
           then
             let
               rootDir = lib.replaceStrings [ "/*/" "/*" ] [ "" "" ] member;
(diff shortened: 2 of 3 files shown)


config.nix
```diff
@@ -258,17 +258,22 @@ let
         # this turns members like "foo/*" into [ "foo/bar" "foo/baz" ]
         # as in https://github.com/rust-analyzer/rust-analyzer/blob/b2ed130ffd9c79de26249a1dfb2a8312d6af12b3/Cargo.toml#L2
         expandMember = member:
-          if lib.hasSuffix "/*" member
+          if lib.hasInfix "*" member
```
nmattia: what about `hasSuffix "/*" || hasSuffix "/*/"`? This is a tad misleading, I first assumed it handled cases like `foo/*/bar` expanding to `foo/quux/bar`.

config.nix
```diff
@@ -258,17 +258,22 @@ let
...
           then
             let
-              rootDir = lib.removeSuffix "/*" member;
-              subdirs = builtins.attrNames (builtins.readDir (root + "/${rootDir}"));
-            in
-              map (subdir: "${rootDir}/${subdir}") subdirs
+              rootDir = lib.replaceStrings [ "/*/" "/*" ] [ "" "" ] member;
+              subdirs = (
+                builtins.attrNames (
+                  lib.filterAttrs
```
nmattia: nice!

config.nix
sean-bennett112: Oh, good catch, that makes sense.

sean-bennett112:
Updated to use `lib.hasSuffix` and enabled github actions on the fork.

sean-bennett112:
Hmm, I might be missing something(this is my first time using Github Actions), but it looks like there's a failure in the post action due to a missing environment variable:

`/home/runner/work/naersk/naersk/.github/actions/nix/install-nix: line 71: CACHIX_SIGNING_KEY: unbound variable`

sean-bennett112:
Hmm, well I'm not sure why the checks are not showing up in this pull request, but the actions now [pass](https://github.com/sean-bennett112/naersk/runs/519975355) on the fork. I had to make a small fix to the action to make sure `CACHIX_SIGNING_KEY` was set as well.

.github/actions/nix/install-nix
```diff
@@ -62,7 +62,7 @@ save_cache() {
     echo "Using cache '$cachix_cache' for '$GITHUB_REPOSITORY'"
 
     set +x
-    if [ -z "${CACHIX_SIGNING_KEY:-}" ] && [ -n "${INPUT_CACHIX_SIGNING_KEY:-}" ]; then
+    if [ -z "${CACHIX_SIGNING_KEY:=}" ] && [ -n "${INPUT_CACHIX_SIGNING_KEY:-}" ]; then
```
nmattia: oh, good catch!

PR #87, opened 2020-03-23
Resolve #48

### LINKED ISSUE #48: Parse lockfile for expanded revision when toml has shortrev
https://github.com/pingcap/talent-plan/blob/dbd711344ff455f2d7942a494a0b97d0a6a6943c/rust/projects/project-4/Cargo.lock

^ The `cargo.toml` only stores the short rev (`8cc906b`) which Nix chokes on. The full revision is available in the lock file, we could grab it from there.

bqv:
I believe this should resolve the issue of shortrevs being passed to fetchgit (#48)

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -76,6 +76,7 @@ rec
   #   }
   findGitDependencies =
     { cargotomls
+    , cargolock
     }:
       let
         tomlDependencies = cargotoml:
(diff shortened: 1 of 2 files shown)


nmattia:
Thanks a lot! Could you add a brief comment explaining what's happening? I lost context :/

nmattia:
Ah, got it. Looks good! I'll run the tests locally. Alternatively you can enable GitHub actions on your fork.

nmattia:
I'm getting an error:

```
cargo -Z unstable-options --locked build $cargo_release -j "$NIX_BUILD_CORES" --out-dir out --message-format=$cargo_message_format
error: no matching package named `rand` found
location searched: https://github.com/rust-random/rand?rev=5b0d87436999c3d930e17f9f10bdb68cd0aa38df#5b0d8743
prerelease package needs to be specified explicitly
rand = { version = "0.7.0-pre.1" }
required by package `crate-b v0.1.0 (/build/dummy-src/crate-b)`
[naersk] cargo returned with exit code 101, exiting
```

When running:

```
$ nix-build ./test.nix -A git-dep-dup
```

bqv:
Seems actions are already enabled!

Otherwise, good catch, seems I'd not considered that case and the lock matching wasn't considering the specific shortrevs. The commit I've just pushed fixes that, and the tests now all pass locally, but I'll squash it in a second.

If not already clear, all this does is update the crate parsing so that the shortrevs get replaced by full revision hashes where necessary, so fetchgit doesn't crash out.

lib.nix
```diff
@@ -92,7 +93,14 @@ rec
                   checkout = builtins.fetchGit ({
                     url = v.git;
                   } // lib.optionalAttrs (v ? rev) {
-                    rev = v.rev;
+                    rev = let
+                            query = p: p.name == k && (lib.substring 0 (4 + lib.stringLength v.git) p.source) == "git+${v.git}";
+                            extractRevision = url: lib.last (lib.splitString "#" url);
```
nmattia: nitpick: for future proofing we could grab anything between `#` and `&`, but this is good enough for now!

Issue #110, opened 2020-06-20
Git submodules not being initialized?


Hoverbear:
Hi!

I'm trying to use this tool to build https://github.com/timberio/vector/commit/0836bcdc0c63e35061c003079edff06af6e3f198 which depends on https://github.com/bytecodealliance/lucet/commit/d741e2e30df19d1f8abd1a8660219ac494336b76 

You can see here, one of the subcrates of `lucet`, `lucet-wasi`, depends on `wasi-common` which is a git submodule in the root. 

![image](https://user-images.githubusercontent.com/130903/85203794-f1e6f280-b2c4-11ea-9324-b8bbf2207a3f.png)

https://github.com/bytecodealliance/lucet/blob/d741e2e30df19d1f8abd1a8660219ac494336b76/lucet-wasi/Cargo.toml#L25

![image](https://user-images.githubusercontent.com/130903/85203812-0dea9400-b2c5-11ea-96ab-c77955c0e143.png)

![image](https://user-images.githubusercontent.com/130903/85203827-22c72780-b2c5-11ea-887f-84070457a728.png)

This seems to not be handled. :(

![image](https://user-images.githubusercontent.com/130903/85203667-20180280-b2c4-11ea-8c9d-7f16ae9395ca.png).

You can reproduce this with:

```git
git clone git@github.com:timberio/vector.git
git checkout 0836bcdc0c63e35061c003079edff06af6e3f198
make neu-release-binary-x86_64-unknown-linux-gnu
```

I see https://github.com/NixOS/nix/pull/3166 but haven't managed to invoke it.

![image](https://user-images.githubusercontent.com/130903/85204639-b4856380-b2ca-11ea-9ab1-1a8dba1b3f38.png)

nmattia:
> I see NixOS/nix#3166 but haven't managed to invoke it.

The PR was merged very recently, did you install the latest `nix` executable?

Hoverbear:
It's sadly not on the latest `nixos-unstable` :(

```
vector on  nix-builds [$!?] is 📦 v0.10.0 via ⬢ v12.18.0 via 🦀 v1.44.0 via ❄️  impure took 2s 
❯ nix repl                                                      
Welcome to Nix version 2.3.6. Type :? for help.

nix-repl> builtins.fetchGit { url = ""; submodules = true; }
error: unsupported argument 'submodules' to 'fetchGit', at (string):1:31
```

This is the latest: https://github.com/NixOS/nix/releases/tag/2.3.6

[...]

PR #111, opened 2020-06-20
Add test git-dep-with-submodule

Hoverbear:
A reproduction test and working on a fix for #110.

Reproduce with

```bash
nix build -f test.nix git-dep-with-submodule
```


test.nix
```diff
@@ -121,6 +121,12 @@ let
         cargoOptions = (opts: opts ++ [ "--locked" ]);
       };
 
+      git-dep-with-submodule = naersk.buildPackage {
+        doCheck = true;
+        src = ./test/git-dep-with-submodule;
```
cole-h, line 126: I think you forgot to check in the testcase itself.

Hoverbear: Yup I did! 🤦‍♀️ 

Good news  is this new commit includes a fix!

Hoverbear:
So I included a ... partial fix? I'm not sure. I don't think moving to `pkgs.fetchgit` is better. Unfortunately `submodules = true;` isn't supported on `fetchGit`... Do you have any advice?

lib.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, writeText, runCommand, remarshal }:
+{ lib, writeText, runCommand, remarshal, fetchgit }:
```
nmattia, line 1: is `fetchgit` actually used?

Hoverbear: So I actually got stuck here, since `fetchgit` requires `sha256` and as far as I can tell we don't have that! The `builtins.fetchGit` doesn't, but it's not due for release until Nix 2.4 (Domen told me!)

I could use your guidance or assistance. :)

Hoverbear:
I don't have a working solution to this so I'll close and leave the report open. :)

nmattia:
Hey! I couldn't give this more attention, I have a lot on my plate at the moment. I'll try to look https://github.com/nmattia/naersk/issues/110 as soon as possible. In the meantime, thanks for giving this a shot!

Hoverbear:
I think we all do right now. :) Keep up your really amazing work @nmattia ! Thanks for looking at this!

PR #119, opened 2020-10-08
Make top-level override easier

### LINKED ISSUE #105: How to specify a hook (e.g. postInstall) only for the main package?
I was wondering if it's possible (or could be made possible) to keep extraneous top-level directories. For example, my project has a `completions` folder that I would like available, so I can add `installShellCompletion` to my `postInstall`. Currently, I do `installShellCompletion --fish ${./completions/passrs.fish}`, but it would be nice if I could instead just do `installShellCompletion --fish $src/completions/passrs.fish`.

nmattia:
This adds `overrideMain f` which is equivalent to calling
`drv.overrideAttrs f`.

Closes #105 
CC @cole-h


cole-h (approved):
Work perfectly, thank you! :heart:

PR #125, opened 2020-11-21
Improve nix flake templates

jD91mZM2:
I noticed that the current templates aren't perfect - they use `pkgs.callPackage naersk {}` which is pointless since naersk has a `flake.nix`. I also added a template that supports a custom rust version, in case someone needs that


nmattia (dismissed):
This is great, thanks a lot!

PR #129, opened 2020-11-30
port wasm test to 20.09 and unstable

petabyteboy:
Depends on https://github.com/nmattia/naersk/pull/128

Built muslTests with nixpkgs-20.09, but they also evaluate with nixpkgs and nixpkgs-20.03.

PR #128, opened 2020-11-29
update test environment

petabyteboy:
I hope this doesn't break your cachix integration. I didn't take a closer look at how that works, but with cachix/install-nix-action the workflows work in my fork.

PR #130, opened 2020-11-30
add musl test

petabyteboy:
Depends on #128, #129


nmattia:
Love it, let's figure out what to do about #128 and #129 then I'm more than happy to merge this!

nmattia:
Nice, thanks a lot!

PR #131, opened 2020-11-30
use cachix/install-nix-action in github workflow

petabyteboy:
The nix github action included in the naersk repository is broken in several ways:

- Nix installer fails with --extra-nix-conf
- GitHub removed some functionality like the `echo "::...` commands.
- even after fixing all this it does not use binary cache for some reason?

cachix/install-nix-action + cachix/cachix-action take about 8s on linux compared to 7s for the included action


nmattia:
Hey, thanks for trying to fix the action. I'll need to have another look at the cachix action, I'm hoping to have some time tomorrow. I tried fixing it as well before I saw #132 and this; indeed it seems somewhat complicated...

petabyteboy:
Any news?

blitz:
@nmattia Any blockers here?

nmattia:
Sorry guys, just a lot on my plate at the moment. Will look at this today.

Issue #133, opened 2020-11-30
Cannot detect local path dependencies


Rizary:
I have this dependency in my `Cargo.toml`:
```
[package]
name = "rust-backend"
version = "0.1.0"
authors = ["Andika Demas Riyandi <andika.riyan@gmail.com>"]
edition = "2018"
build = "build.rs"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
...
common = { path = "../common/", optional=false }
...

[build-dependencies]
ructe = { version = "0.12", features = ["sass", "tide013"] }
```

Since my `common` package is a local library, when I build my `rust-backend` I got the following error:
```
error: --- Error --------------------------------------------------------------------------------------------------- nix
builder for '/nix/store/mqg5569yzxxdxw8nfvq920lv41gpzwi3-rust-backend-deps-0.1.0.drv' failed with exit code 101; last 10 log lines:

  Caused by:
    Unable to update /tmp/nix-build-rust-backend-deps-0.1.0.drv-0/common

  Caused by:
    failed to read `/tmp/nix-build-rust-backend-deps-0.1.0.drv-0/common/Cargo.toml`

  Caused by:
    No such file or directory (os error 2)
  [naersk] cargo returned with exit code 101, exiting
error: --- Error --------------------------------------------------------------------------------------------------- nix
```

I tried with `rustPlatform.buildRustPackage` and the behavior is the same. is there something wrong with my setup?

You can find the project here: https://github.com/numtide/todomvc-nix/tree/flake

goertzenator:
I also ran into a failing path dependency:

```
[dependencies]
clap = "2.32"
libc = "0.2.81"
...
usbfs = {path = "usbfs-rs", version = "0.1"}
```

I see that there are multi-crate workspace tests for naersk but the crates are not depending on each other.  Adding such a dependency would be a good way to observe this issue.

Ekleog:
FWIW, within a workspace local path dependencies work without problems for me. What doesn't work is local path dependencies *outside of* workspaces.

In my case, I hit the issue while trying to build `cargo` itself with `naersk`: `cargo` is architectured as one root crate and then sub-crates in the same repo, without a top-level workspace crate, and thus naersk fails building it due to this reason.

While I see the general problem (especially `../` path dependencies) as (very?) hard to solve, as it requires fetching additional paths, IMO this problem could be solved for the “usual” case where the other crates are already inside `$src`.

[...]

PR #135, opened 2020-12-31
Also copy local path-dependencies

Ekleog:
Hello!

I'm trying to use naersk to build a patched `cargo` from source, and in the process of doing so I hit #133.

Here is a PR that partially fixes the issue, and allows setting `copySources` to the dependencies that might still be missed by this non-recursive algorithm.

What do you think about this? It has as drawbacks that path-dependency crates will be compiled twice (once in the “prebuild the deps” phase and once in the “build the code” phase), and also that only direct path-dependencies are (currently) being handled.

What do you think about this approach to #133? Should I try to improve on it, or just to make `copySources` available?

Anyway, thank you for naersk, I'm using it only since recently but it feels really nice to use :)

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -303,21 +303,33 @@ let
       in
         lib.unique (lib.concatMap expandMember listedMembers);
 
-    patchedSources =
+    copySourcesFrom = src;
+
+    copySources =
       let
-        mkRelative = po:
+        chkRelative = po:
           if lib.hasPrefix "/" po.path
           then throw "'${toString src}/Cargo.toml' contains the absolute path '${toString po.path}' which is not allowed under a [patch] section by naersk. Please make it relative to '${toString src}'"
-          else src + "/" + po.path;
+          else po.path;
       in
+        arg.copySources or []
+      ++
         lib.optionals (builtins.hasAttr "patch" toplevelCargotoml)
           (
-            map mkRelative
+            map chkRelative
               (
                 lib.collect (as: lib.isAttrs as && builtins.hasAttr "path" as)
                   toplevelCargotoml.patch
               )
-          );
+          )
+      ++
+        # TODO: this should probably recurse into these other Cargo.toml's? In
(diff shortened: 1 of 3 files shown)


Ekleog:
Coming back to my code from yesterday night, I see four ways forward with this diff:
1. Keep both the path auto-detection logic and the `copySources` escape hatch, in which case we can land this PR as is and then try to improve on the path auto-detection logic
2. Keep the path auto-detection logic but not the `copySources` escape hatch, in which case I'd have to improve on the path auto-detection logic before landing
3. Keep the `copySources` escape hatch but not the path auto-detection logic, in which case I'd have to remove the path auto-detection logic before landing
4. Keep none of those features and explicitly state that naersk doesn't support path dependencies and there is no escape hatch (though that'd make me sad)

Is there one that'd have your favor?

Ekleog:
Hmmmmm using more my fork of naersk with these changes, it looks like the current `path` logic may also happen to detect path dependencies in workspaces, thus losing all the benefits of two-step compilation (though I'm kind of surprised, as it's supposed to only use the root cargo.toml… but maybe this function is called multiple times for virtual manifests? not sure, haven't investigated more yet).

I thus feel like this road may be more full of ambush than the `copySource` escape hatch, especially as the current implementation of auto-following path dependencies is already broken as in it doesn't recursively follow all paths.

So I'd tend towards just adding `copySource` as an argument for now, documenting that it's the expected way to handle non-workspaced path dependencies, and eventually coming back to handling path dependencies actually correctly.

What do you think about this?

Ekleog:
I'm going to close this as I've just pushed https://github.com/nmattia/naersk/pull/136 that does not have the “breaks caching” behavior — it may make sense to come back to the fuller implementation of path-dependency handling, but as “expose `copySources`” and “handle path dependencies outside of workspaces” are two mostly independent changes, they'd most likely better not be mixed together in a single PR

PR #136, opened 2021-01-03
Expose copySource for dealing with path-based dependencies

Ekleog:
This is an alternative to https://github.com/nmattia/naersk/pull/135 ; that doesn't try to implement path dependency parsing, as:
1. it'd need to be done recursively, which requires more work
2. it'd have to make sure that workspaces with path dependencies don't end up breaking most if not all the benefits of two-derivation caching
3. in most cases manually filling in `copySources` should not be too bad a thing to do (eg. to compile `cargo` I just had to set `copySources = ["crates"]`), and anyway to get more of the benefit of two-derivation caching there would be very hard (impossible?)

What do you think about this PR?

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -303,13 +303,20 @@ let
       in
         lib.unique (lib.concatMap expandMember listedMembers);
 
-    patchedSources =
+    # If `copySourcesFrom` is set, then it looks like the benefits brought by
+    # two-step caching break, for unclear reasons as of now. As such, do not set
+    # `copySourcesFrom` if there is no source to actually copy from.
+    copySourcesFrom = if copySources != [] then src else null;
+
+    copySources =
       let
         mkRelative = po:
           if lib.hasPrefix "/" po.path
           then throw "'${toString src}/Cargo.toml' contains the absolute path '${toString po.path}' which is not allowed under a [patch] section by naersk. Please make it relative to '${toString src}'"
-          else src + "/" + po.path;
+          else po.path;
       in
+        arg.copySources or []
+      ++
         lib.optionals (builtins.hasAttr "patch" toplevelCargotoml)
           (
             map mkRelative
--- a/lib.nix
+++ b/lib.nix
@@ -116,7 +116,8 @@ rec
     { cargoconfig   # string
     , cargotomls   # attrset
     , cargolock   # attrset
-    , patchedSources # list of paths that should be copied to the output
+    , copySources # list of paths that should be copied to the output
+    , copySourcesFrom # path from which to copy ${copySources}
     }:
       let
         config = writeText "config" cargoconfig;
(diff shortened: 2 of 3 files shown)


lib.nix
```diff
@@ -142,15 +143,17 @@ rec
...
             mkdir -p $out/.cargo
             ${lib.optionalString (! isNull cargoconfig) "cp ${config} $out/.cargo/config"}
             cp ${cargolock'} $out/Cargo.lock
 
-            for p in $patchedSources; do
+            for p in $copySources; do
               echo "Copying patched source $p to $out..."
-              cp -R "$p" "$out/"
+              mkdir -p "$out/$p"
+              rmdir "$out/$p"
```
nmattia, line 155: Why create the directory then remove it?

config.nix
```diff
@@ -303,13 +303,17 @@ let
       in
         lib.unique (lib.concatMap expandMember listedMembers);
 
-    patchedSources =
+    copySourcesFrom = if copySources != [] then src else null;
```
nmattia, line 306: Why not always pass `src` here?

nmattia (commented):
I like the idea! It's indeed simpler. I'll need to look into the implementation a bit more seriously but I already have a few questions.

Thanks!

lib.nix
Ekleog: The trick is `mkdir -p` creates all parent directories, so what this actually does is create all directories up to the parent of `$p`. An alternative would be `mkdir -p "$out/$(dirname "$p")"`, would that look better to you? (it may behave differently for paths with `..`, but these would probably be broken anyway, so…)

config.nix
Ekleog: I originally always passed `src`, but it turns out it would add a dependency on `src` all the time, which breaks the two-step compilation process and forces all dependencies to always be rebuilt.

Actually, this is probably also an issue in the implementation for path-based patch dependencies things, that also use `patchedSources`: as soon as there's a patch the deps-dev derivation depends on the src derivation, which breaks deps-dev caching. Not having had any idea to solve this issue, I just made sure there's no regression when not passing in `copySources`.

Should I eg. add a comment about that here?

lib.nix
nmattia: Ah I see! Either works, in both cases would be great to have a one-liner (`mkdir -p "$out/$p" && rmdir "$out/$p"`) with a comment explaining what's happening. Thanks!

config.nix
nmattia: Yes that would be great! Still wrapping my head around the caching issue.

Ekleog:
I've just pushed the missing comments asked in reviews :)

nmattia (approved):
Thanks a lot!

PR #140, opened 2021-01-14
make fetchgit work when not fetching from master

Ekleog:
This is required so that fetching dependencies from a feature branch is
possible.

See also https://github.com/NixOS/nix/pull/3408

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -108,6 +108,7 @@ rec
                     url = val.git;
                   } // lib.optionalAttrs (val ? rev) {
                     rev = val.rev;
+                    ref = val.rev;
                   } // lib.optionalAttrs (val ? branch) {
                     ref = val.branch;
                   } // lib.optionalAttrs (val ? tag) {


nmattia (commented):
Thanks for the PR, can you explain how this works? Is the `ref` passed as-is to git, or does Nix do anything special in that case?

Ekleog:
I don't know that much about `fetchGit` either, but according to the nix manual `ref` is:
> The git ref to look for the requested revision under. This is often a branch or tag name. Defaults to HEAD.
>
> By default, the ref value is prefixed with refs/heads/. As of Nix 2.3.0 Nix will not prefix refs/heads/ if ref starts with refs/. 

In practice, at least for GitHub, doing this allows fetching from a commit sha that is not in the direct history of master, while without it nix fails fetching the git repo

nmattia:
@Ekleog I fixed some conflicts, does this LGTY?

Ekleog:
It does, thank you! I've just force-pushed the branch by squashing this commit in and removing the stray newline that had made it through, so that git history can be cleaner :)

nmattia:
Thanks a lot!

Ekleog:
Thank *you* for maintaining naersk! :D

archseer:
Hmm I'm not sure if this fix is entirely correct, I'm getting

```
fetching Git repository 'https://github.com/fdehau/tui-rs'fatal: couldn't find remote ref refs/heads/8832281dcf1756cfbdfad6c56b969f5f1dafcaa5
```

For a dependency I manually pin to a specific master branch commit:

```
tui = { git = "https://github.com/fdehau/tui-rs", default-features = false, features = ["crossterm"] }
```

Mic92:
Right. Reverting it fixes the issue: https://github.com/nmattia/naersk/issues/149

Ekleog:
Thank you for the debugging and sorry about that! I'm sad to see that it looks like there is no one way to just fetch a revision from github, and I must say I don't understand when one version is supposed to work and when the other version is :/

VanCoding:
I recently tried to package [prisma-engines](https://github.com/prisma/prisma-engines), but naersk was unable to deal with the git revs in the [Cargo.lock](https://github.com/prisma/prisma-engines/blob/master/Cargo.lock#L801).

There, the revs are specified with a `#`, but they don't seem to be detected. Has this something to do with that?

PR #141, opened 2021-01-30
Support flake-based workspaces

jD91mZM2:
Flake sources are nix derivations, which means they automatically get made read-only. This causes overwriting Cargo.toml and potentially build.rs fail. The cure to this is to either chmod the sources after copy, or in this case copy without preserving the file's original mode.


nmattia (approved):
oh thanks!

PR #147, opened 2021-02-10
Use IFD to reduce copySources invalidation surface

exFalso:
This PR fixes `copySources` invalidating the two-stage build. The reasons for the invalidation:

* `copySourcesFrom` was an attribute of the `dummy-src` derivation, causing invalidation if any sources changed
* `copySourcesFrom` was used directly as the cp source directory with `$copySourcesFrom/$p`, invalidating the command script even if `$p` has not changed

The PR introduces the following changes:
* `$copySourcesFrom/$p` changed to `${copySourcesFrom + "/${p}"}`, and the surrounding for loop to a `concatStrings`. This causes the subdirectory to be treated like a proper nix path, so as long as `copySourcesFrom` is a local path, the script will only invalidate if `$p` changes.
* An IFD indirection was introduced in the form of `ifdForceCopy`. This is to handle cases where `copySourcesFrom` is *not* a local path, but a /nix/store one (like in our use case). The indirection effectively "snips" the dependency chain, so it doesn't matter what derivation `copySourcesFrom` comes from and how it invalidates, `ifdForceCopy` will re-copy the subdirectory as a separate "leaf" store path. NOTE: this solution will effectively re-copy the sources **twice**. This could be reduced somewhat by using gzip instead of straight copying for the IFD.
* `copySourcesFrom` was removed from `dummy-src`'s attributes

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -149,21 +149,28 @@ rec
           (k: v: "${k}:${builtinz.writeTOML "Cargo.toml" (fixupCargoToml v)}")
           cargotomls;
 
+      ifdForceCopy = path:
+        let
+          derivation = runCommand "ifd-force-copy" {} ''
+            mkdir -p $out
+            echo "./path" > $out/default.nix
+            cp --no-preserve=mode -R ${path} $out/path
+          '';
+        in import "${derivation}";
+
       in
         runCommand "dummy-src"
-          { inherit copySources copySourcesFrom cargotomlss; }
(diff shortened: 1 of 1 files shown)


Patryk27:
Hi, thanks for this merge request - unfortunately, unless there were any recent developments in this area, to keep naersk working with Hydra, we have to avoid using IFD; and so given the choice, I'd rather keep the current approach; thanks again!

PR #151, opened 2021-02-16
fix(builtins/to-toml): support int type

cole-h:
Fixes https://github.com/nmattia/naersk/issues/148.

"As of rust-lang/cargo@7dd9872, Cargo adds a new `version = 3` key to
the lockfile's top level"


nmattia (approved):
Nice, thanks!

PR #160, opened 2021-03-25
Fix windows executables not having .exe suffix.

gcoakes:


nmattia (approved):
Yep, that makes sense. Thanks!

PR #161, opened 2021-03-25
Add examples and templates for cross compilation.

gcoakes:
This adds examples and templates for both Windows (mingw-w64) and static musl using flakes. A project can be initialized with `nix flake init -t github:nmattia/naersk#cross-windows` or similar for musl.

I can imagine the same being done for other possible cross targets (i.e.: wasm & MacOS), but for now, I just implemented the ones that were of interest to me.


nmattia (dismissed):
That's great, thanks a lot! Would you mind adding a sentence or two in the README about this? Otherwise I'm afraid people will never find it!

nmattia (approved):
This is great, thanks!

PR #179, opened 2021-06-08
ci: run all tests on master

zimbatm:
Some tests like the wasm one were never executed because they are
neither --fast, nor tested in the main CI.

This copies the fast.yml into test.yml, minus the --fast flag.


nmattia (approved):
Thanks, nice brainfart on my end.

PR #189, opened 2021-08-19
Don't rely on unstable flags.

### LINKED ISSUE #100: Is it possible for naersk to use stable Rust
Hi apologies for my naïvety. I'm very new to Nix.

All the instructions I have found so far describe using nightly rust with naersk. I would rather use a release version of stable rust. Is this possible?

I tried removing `"-Z" "unstable-options"` from `cargoOptions` but it seems that was there to allow the use of `--out-dir`. Can the [CARGO_TARGET_DIR](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads) be used instead?

dpc:
Fix #100

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -28,18 +28,18 @@ let
     # The command to use for the build.
     cargoBuild =
       allowFun attrs0 "cargoBuild"
-        ''cargo $cargo_options build $cargo_build_options >> $cargo_build_output_json'';
+        ''env CARGO_TARGET_DIR=out cargo $cargo_options build $cargo_build_options >> $cargo_build_output_json'';
 
     # Options passed to cargo build, i.e. `cargo build <OPTS>`. These options
     # can be accessed during the build through the environment variable
     # `cargo_build_options`. <br/>
-    # Note: naersk relies on the `--out-dir out` option and the
+    # Note: naersk relies on the the
     # `--message-format` option. The `$cargo_message_format` variable is set
     # based on the cargo version.<br/>
     # Note: these values are not (shell) escaped, meaning that you can use
     # environment variables but must be careful when introducing e.g. spaces. <br/>
     cargoBuildOptions =
-      allowFun attrs0 "cargoBuildOptions" [ "$cargo_release" ''-j "$NIX_BUILD_CORES"'' "--out-dir" "out" "--message-format=$cargo_message_format" ];
+      allowFun attrs0 "cargoBuildOptions" [ "$cargo_release" ''-j "$NIX_BUILD_CORES"'' "--message-format=$cargo_message_format" ];
 
 
     # When `true`, rustc remaps the (`/nix/store`) source paths to `/sources`
@@ -71,7 +71,7 @@ let
     # Note: these values are not (shell) escaped, meaning that you can use
     # environment variables but must be careful when introducing e.g. spaces. <br/>
     cargoOptions =
-      allowFun attrs0 "cargoOptions" [ "-Z" "unstable-options" ];
+      allowFun attrs0 "cargoOptions" [ ];
 
     # When true, `cargo doc` is run and a new output `doc` is generated.
     doDoc = attrs0.doDoc or false;


dpc:
I tested it with `beta` and `stable` and on my little project seems to work. Other than this I have no idea what I'm doing.

nmattia:
Thanks! @basvandijk do you remember why we used `--out-dir` and not `CARGO_TARGET_DIR`? I know there was a reason, but right it escapes me...

nmattia:
Ah, I think I knew. The difference is explained here: https://github.com/rust-lang/cargo/pull/5203

The `--out-dir` parameter is only for final artifacts, and not intermediary results like `rlibs` (I think). Can you check the final build size with and without your changes?

dpc:
```
CARGO_TARGET_DIR — Location of where to place all generated artifacts, relative to the current working directory. See build.target-dir to set via config.
```

https://doc.rust-lang.org/cargo/reference/environment-variables.html

Is it a problem if everything goes into the out dir?

dpc:
Would it be OK if we just delete needless stuff from the output dir? I mean - it's not all that difficult. All the subdirs, and then bunch of `.d` and `.rmeta` files, and job done?

dpc:
BTW. I have no idea why, but I'm building my Rust project under a special name (just to make sure I am looking at the right output):

```
~/l/dotr (master)> find /nix/store/ | grep dotrx
/nix/store/f6x9qjq27ixjdlb38g1fdc95rc7157vr-dotr-0.4.0/bin/dotrx
```

No other files than the binary. No idea why.

nmattia:
Mh ok that's odd. Maybe that actually works since the top-level derivation only builds the executable? what if you set `single-step` to `true` so that everything is built in one derivation?

dpc:
Weird. Same? I've changed the name to `dotrxx` now.

```
[I] 08-20 11:41 dpc@futex ~/l/dotr (master)> find /nix/store/y9r0lhv3q8mkk3vxlrisjy38c1dxwmlx-dotr-0.4.0
/nix/store/y9r0lhv3q8mkk3vxlrisjy38c1dxwmlx-dotr-0.4.0
/nix/store/y9r0lhv3q8mkk3vxlrisjy38c1dxwmlx-dotr-0.4.0/bin
/nix/store/y9r0lhv3q8mkk3vxlrisjy38c1dxwmlx-dotr-0.4.0/bin/dotrxx
```

previous one (no singleStep)

```
[I] 08-20 11:41 dpc@futex ~/l/dotr (master)> find /nix/store/f6x9qjq27ixjdlb38g1fdc95rc7157vr-dotr-0.4.0
/nix/store/f6x9qjq27ixjdlb38g1fdc95rc7157vr-dotr-0.4.0
/nix/store/f6x9qjq27ixjdlb38g1fdc95rc7157vr-dotr-0.4.0/bin
/nix/store/f6x9qjq27ixjdlb38g1fdc95rc7157vr-dotr-0.4.0/bin/dotrx
```

dpc:
Is it because of 

```
    # The value is written to the `cargo_bins_jq_filter` variable.                                                                           
    copyBinsFilter = attrs0.copyBinsFilter or                                                                                                
      ''select(.reason == "compiler-artifact" and .executable != null and .profile.test == false)'';                                         
 ```
 
 ?

nmattia:
Well then! if it breaks anything, we'll revert and add a test :)

nmattia:
Actually I just reverted the commit. I realized that you were setting `CARGO_TARGET_DIR` locally, for the build only, meaning that cargo test, clippy, doc, etc will be rebuilding from scratch, since they'd revert to the default `target/`!

dpc:
Does it mean I should add that flag into other places as well?

nmattia:
Yes I think you can export it like so: https://github.com/nix-community/naersk/blob/df71f5e4babda41cd919a8684b72218e2e809fa9/build.nix#L241

Thanks!

dpc:
I am completely confused what am I doing. I don't understand why exactly it was reverted and what is the problem in the first place. :)

PR #191, opened 2021-08-29
Don't rely on unstable flags

dpc:
Another attempt at #100

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -239,6 +239,7 @@ let
       # threads as there are cores. This is often too much parallelism so we
       # reduce it to $NIX_BUILD_CORES if not specified by the caller.
       export RUST_TEST_THREADS="''${RUST_TEST_THREADS:-$NIX_BUILD_CORES}"
+      export CARGO_TARGET_DIR="out"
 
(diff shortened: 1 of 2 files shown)


veehaitch (dismissed):
Works like a charm!

build.nix
```diff
@@ -239,6 +239,7 @@ let
       # threads as there are cores. This is often too much parallelism so we
       # reduce it to $NIX_BUILD_CORES if not specified by the caller.
       export RUST_TEST_THREADS="''${RUST_TEST_THREADS:-$NIX_BUILD_CORES}"
+      export CARGO_TARGET_DIR="$out"
```
nmattia, line 242: I think you'll want `="out"` as opposed to `$out`! we want the directory `out`, not the build's output dir. Binaries are _then_ copied from `out` to `$out/bin`.

nmattia (commented):
Thanks a lot! Just came back from holidays, sorry for the late review.

build.nix
dpc: Done.

nmattia:
Alright let's try!

PR #201, opened 2021-11-10
Fix doDoc

blitz:
Somewhere since commit e09c320446c5c2516d430803f7b19f5833781337, naersk is unable to handle `doDoc`. It seems the code around the cargo doc invocation assumes that documentation appears in target/doc, while it is actually written to out/doc now.

If you need a tiny test case, I have the old working and current failing versions of naersk ready to try [here](https://github.com/blitz/naersk-doc-bug-example).

I've seen that there is actually a test for `doDoc` but the pipeline is red since June. I could help with fixing this, if someone has some pointers what to actually fix. :)

PR #203, opened 2021-11-11
Fix copying cargo artifacts from deps

### LINKED ISSUE #202: Cargo artifacts not being copied from deps derivation
I noticed that building my crate via naersk was resulting in cargo rebuilding the entire dependency tree from scratch every time, even when the Cargo.lock had not changed.

Upon closer inspection I discovered that the deps derivation results in a `target.tar.zstd` which unarchives to an empty directory.

I think the culprit may be [forcing cargo to use `out` as its target directory](https://github.com/nix-community/naersk/blob/074d81b1a45145f076b2adf93184073fc9615397/build.nix#L243) but then [copying the artifacts from the (unused) `target` directory](https://github.com/nix-community/naersk/blob/074d81b1a45145f076b2adf93184073fc9615397/build.nix#L402-L415).

ipetkov:
* Previously the CARGO_TARGET_DIR environment was set to the `out`
  directory which resulted in cargo placing all of its artifacts there
  instead of the default `target` directory
* Later on, the build scripts would then copy the (empty) `target`
  directory to `$out` resulting in cargo rebuilding the entire
  dependency tree from scratch whenever the top level crate's source was
  changed
* Now we omit the CARGO_TARGET_DIR environment variable and inspect the
  `target` directory for any binaries instead of the `out` directory

Fixes #202

### DIFF
--- a/test.nix
+++ b/test.nix
@@ -97,9 +97,56 @@ let
           touch $out
         '';
 
+      # Tests that the builtDependencies derivation can successfully be unpacked
+      # and that it actually contains cargo's output artifacts. If the result is
+      # ever empty, cargo will still succeed in building the top level crate, except
+      # it will need to rebuild all dependencies from scratch, which is wasteful.
+      # See https://github.com/nix-community/naersk/issues/202
+      depsTargetNotEmpty = pkgs.runCommand "depsTargetNotEmpty"
+        { inherit (simple-dep) builtDependencies; }
+        ''
+          for dep in $builtDependencies; do
+            # Make destination directory for unarchiving
+            mkdir dst
+            ${pkgs.zstd}/bin/zstd -d "$dep/target.tar.zst" --stdout | tar -x -C ./dst
+
+            if [ -z "$(ls -A ./dst)" ]; then
+              echo target directory is empty: "$dep"
+              return 1
+            fi
+
+            rm -rf ./dst
+          done
+
+          # Success
+          touch $out
+        '';
+
+      # Same as the test above except checks when the builtDependencies
+      # derivation is not compressed.
+      depsUncompressedTargetNotEmpty = pkgs.runCommand "depsUncompressedTargetNotEmpty"
+        { inherit (simple-dep-no-compress) builtDependencies; }
+        ''
+          for dep in $builtDependencies; do
+            if [ -z "$(ls -A "$dep"/target)" ]; then
+              echo target directory is empty: "$dep"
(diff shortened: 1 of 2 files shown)


nmattia:
Thanks for the PR! I fixed part of the CI in #204 . There's one issue remaining: #205

Re. your PR, how does it relate to https://github.com/nix-community/naersk/pull/201?

ipetkov:
> Re. your PR, how does it relate to #201?

Ah I hadn't noticed it before submitting this PR. I think they both solve different issues in opposite ways. As blitz commented on the other #201, it shouldn't be needed if we merge #203

IMO #203 is a more general solution, as the next person to come along and make a change can expect cargo artifacts to be in `target` instead of needing to remember to use `out` everywhere (while also avoiding getting confused with `$out`)

nmattia (approved):
Mostly rubbing stamping here, there's been so many changes around how we copy the executables that I can't keep track 😂 

@blitz I closed #201 in favor of this, would you mind having a look and approving if it looks good?

test.nix
```diff
@@ -97,11 +97,51 @@ let
           touch $out
         '';
 
+      depsTargetNotEmpty = pkgs.runCommand "depsTargetNotEmpty"
```
blitz, line 100: nit: It would help if there is a small comment what these tests actually test for.

blitz (approved):
This looks good! It also fixes `doDoc = true`.

I've run the test suite manually and it reports success.

test.nix
ipetkov: Done!

blitz (approved):
Thanks for the explanation in the tests. Much easier to understand now.

blitz:
@nmattia Ping. This is ready to be merged. :)

blitz:
@anderssorby can you take a look?

Anderssorby:
@blitz Looks good to me. This is a much needed fix.

blitz:
@nmattia has to merge this I guess. At least I can't.

ipetkov:
Might be good to update the CODEOWNERS file with the new maintainers so we don't have to keep spamming nmattia :)

(If someone with write access wants to merge the commit and push directly that's cool too)

blitz:
> Might be good to update the CODEOWNERS file with the new maintainers so we don't have to keep spamming nmattia :)

See #208.

nmattia:
I'll let you do the honours :)

PR #204, opened 2021-11-15
Try fixing CI

nmattia:


nmattia:
This actually fixes the CI jobs, excepts for the artifacts upload. We need to use a nix-community cachix and update the secrets. CC @domenkozar maybe?

Issue #206, opened 2021-11-15
Project needs more maintainers!


nmattia:
Hi there! 

This project needs some love. I do not have time to maintain it, and @jonringer is already [snowed under](https://github.com/nix-community/naersk/issues/183#issuecomment-902312399). If any kind soul feels like dedicating some love to this project, please speak up!

blitz:
@nmattia Are you looking for a "full-time" maintainer or do you want to offload day-to-day things like code review and cleanups?

nmattia:
@blitz mostly day-today things like code reviews and merging PRs. If there's any big feature that needs to be implemented then we'll see! but at least someone giving feedback to contributors would be great. :)

jonringer:
Yea, I'm already the release manager for NixOS and already dedicated to many other things. I was a bit optimistic about being able to dedicate more time to this. :(

blitz:
I'm happy to review code and also invest time in cleanups.

Anderssorby:
I've been using this library a lot the last year and I would gladly help some. The docs and README need some updates and there are several features that I think would be nice.

nmattia:
This is very much appreciated! I gave you both write access to the repo. Thanks for you help :)

PR #208, opened 2021-11-20
Update CODEOWNERS

blitz:
Add @Anderssorby and me to the CODEOWNERS file, which will ping us on pull requests. This should also make our reviews count towards the review limit and allow us to actually merge things. :)

Relates to #206.


nmattia:
Thanks :)

Issue #205, opened 2021-11-15
Set up cachix


nmattia:
Before moving to nix-community, naersk was using my personal cachix cache. I removed the secrets; a new cachix cache need to be created and the secrets updated. I do not know how nix-community deals with caches.

CC @domenkozar @jonringer

domenkozar:
And admin needst to create a write key /cc @zimbatm

PR #209, opened 2021-11-22
Repair CI

blitz:
Switch to upstream [install-nix-action](https://github.com/marketplace/actions/install-nix?version=v16) instead of bundled one. This also disables uploading artifacts to cachix and should thus repair the CI. As the upload was not working any how (see #205) we are not losing anything here.


nmattia:
I've been against using install-nix-action in the past because it's a big javascript blob that we need to trust is correct (because compiled from ts). Also I find bash a much more natural language for setting up builds. Finally, the cache should just work... but it looks like I can't since I'm not an admin for naersk.cachix.org? 

CC @domenkozar you might have some thoughts here :)

blitz:
> I've been against using install-nix-action in the past because it's a big javascript blob that we need to trust is correct (because compiled from ts). Also I find bash a much more natural language for setting up builds.

I'm confused. If I see this correctly, install-nix-action is the evolution of the bash script I just deleted. I don't see more Javascript in there: https://github.com/cachix/install-nix-action/blob/master/lib/install-nix.sh

> Finally, the cache should just work... but it looks like I can't since I'm not an admin for naersk.cachix.org?

i've left the signing key alone. If it's configured correctly it should work.

blitz:
@ofborg build onedrive

nmattia:
>  If I see this correctly, install-nix-action is the evolution of the bash script I just deleted.

Ohhh well that's new then, and it's great news :) Let's go then!

> i've left the signing key alone. If it's configured correctly it should work.

Yeah looks like I somehow kicked myself out of the naersk.cachix.org admin group...

Anderssorby (approved):
The install-nix-action seems good to me

Anderssorby:
@nmattia Shall I merge? Do you have some developer chat room btw? Using matrix for example.

nmattia:
Yeah go for it! Does caching work? Otherwise this will lead to a slow and painful death by CI

blitz:
> Do you have some developer chat room btw? Using matrix for example.

I'm js:ukvly.org on Matrix. We can coordinate in #rust:nixos.org as long as it doesn't have a lot of volume.

Issue #149, opened 2021-02-11
Cargo.toml: Git Dependencies


a-kenji:
When I try to build a package with this in the Cargo.toml:
```
[dependencies]
rocket = { git = "https://github.com/SergioBenitez/Rocket.git", rev = "1f1f44f336e5a172361fc1860461bb03667b1ed2", features = ["tls"] }
```

I get this error message:
```
fetching Git repository 'https://github.com/SergioBenitez/Rocket.git'fatal: couldn't find remote ref refs/heads/1f1f44f336e5a172361fc1860461bb03667b1ed2
error: --- Error ------------------------------------------------------------------------------------- nix
program 'git' failed with exit code 128
(use '--show-trace' to show detailed location information)

```
But the rev seems to exist.

If you want I can also attach the --show-trace.

Mic92:
I have a similiar issue. Those are the commands it tries to execute in this case:

```
nix-build        29705  16611    0 /scratch/joerg/.nix-profile/bin/nix-build --show-trace
git              30279  29705    0 /nix/store/cdmxpgjyx8nkxyf64rjvpazpi6p5dslb-git-2.30.0/bin/git -C /scratch/joerg/.cache/nix/gitv3/01lpsaj1cwwnf8a3nh0wglq0jpqc1dazg5864mvw47gdf3nvw8vl cat-file -e 5847f1286492b7191f1400e6647fb220f8941f89
git              30280  29705    0 /nix/store/cdmxpgjyx8nkxyf64rjvpazpi6p5dslb-git-2.30.0/bin/git -C /scratch/joerg/.cache/nix/gitv3/01lpsaj1cwwnf8a3nh0wglq0jpqc1dazg5864mvw47gdf3nvw8vl fetch --quiet --force -- https://github.com/rust-vmm/vm-device refs/heads/5847f1286492b
7191f1400e6647fb220f8941f89:refs/heads/5847f1286492b7191f1400e6647fb220f8941f89
git              30281  30280    0 /nix/store/cdmxpgjyx8nkxyf64rjvpazpi6p5dslb-git-2.30.0/libexec/git-core/git remote-https https://github.com/rust-vmm/vm-device https://github.com/rust-vmm/vm-device
git-remote-http  30282  30281    0 /nix/store/cdmxpgjyx8nkxyf64rjvpazpi6p5dslb-git-2.30.0/libexec/git-core/git-remote-https https://github.com/rust-vmm/vm-device https://github.com/rust-vmm/vm-device
[...]

PR #167, opened 2021-04-15
feat: simplify git dependency detection by only using cargo lock

yusdacra:
This PR also adds `allRefs` and `submodules` support. Since these features were added to `builtins.fetchGit` recently, they are not enabled by default.

PR #211, opened 2021-11-29
tomlDependencies: pass allRefs=true when fetching a `rev`

mhuesch:
this allows us to avoid issues when the default branch is not named
`master`.

I believe this should fix the issues described in:
https://github.com/nix-community/naersk/issues/149
https://github.com/nix-community/naersk/pull/140

using this change seems to fix this error I was seeing:
```
fetching Git repository 'https://github.com/sacredcapital/rep_lang.git'fatal: couldn't find remote ref refs/heads/master
error: program 'git' failed with exit code 128
```
on a repo which had only a `main` branch, and no `master` branch. of note, I was able to work around the issue by creating a dummy `master` branch, but still using the rev on the `main` branch.

using this PR seems to enable a successful build, without requiring the dummy `master` branch which was not actually used.

I reproduced the successful build on multiple machines. however I cannot
be full certain that it works in all cases.

additionally, it seems possible that `allRefs = true` causes annoying
extra fetching which might be noticeable for repos with many refs...

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -148,6 +148,7 @@ rec
                     url = val.git;
                   } // lib.optionalAttrs (val ? rev) {
                     rev = val.rev;
+                    allRefs = true;
                   } // lib.optionalAttrs (val ? branch) {
                     ref = val.branch;
                   } // lib.optionalAttrs (val ? tag) {


mhuesch:
sending this PR in case it is useful. no worries about closing if not :slightly_smiling_face:

nmattia (commented):
Sounds sensible, but I'll let @Anderssorby and @blitz have the final word!

lib.nix
```diff
@@ -148,6 +148,7 @@ rec
                     url = val.git;
                   } // lib.optionalAttrs (val ? rev) {
                     rev = val.rev;
+                    allRefs = true;
```
blitz, line 151: nit: The formatting looks a bit off, but I'm on the GitHub app, so this might just be me.

blitz (approved):
Looks good to me, especially if it fixes pain people actually experience.

Not merging right now, because I can't test it at the moment. (Writing this from my phone.) I'll get back to this in a couple of days or maybe in the meantime @Anderssorby can take a look.

lib.nix
blitz: This was just the github android app UI. Nevermind.

blitz (changes requested):
I didn't realize that `allRefs` is a Nix 2.4 feature.

@mhuesch Is there a way to make this backward compatible to Nix 2.3? Falling back to the old behavior of not specifying the `allRefs` parameter would be fine.

mhuesch:
> I didn't realize that `allRefs` is a Nix 2.4 feature.
> 
> @mhuesch Is there a way to make this backward compatible to Nix 2.3? Falling back to the old behavior of not specifying the `allRefs` parameter would be fine.

ah, neither did I! I am not sure how to do version-conditional code in Nix but I will check into it next week when I get back from the trip I'm on 👍

Anderssorby:
Hm, this may cause unnecessary data to be downloaded which is not desirable. I remember in another project we solved this by specifying the branch as well in the Cargo.toml

mhuesch:
perhaps the best way would be to add an optional attr `allRefs` (which defaults to `false`) to an input attrset somewhere?

my (not very informed) sense is that it'll be tough to do a kind of "granular enabling of allRefs=true", so maybe just a global (to the Cargo project which is pointed to by `src` in naersk's `buildPackage` invocation) enabling? so, adding an optional `allRefs` attr to `buildPackage` input attr set and propagating that down.

the reasoning being that many projects won't have this issue and won't notice the change, but for projects that do, downloading some extra data is preferable to not being able to build :slightly_smiling_face:

jonringer:
The latest stable release bumped the nix version again, if you push to the branch, it should re-trigger CI to run. Which will probably pass now.

jonringer:
21.05 is EOL, may just ask people to pin to a commit before this change if they want to be compatible with nix 2.3

mhuesch:
> The latest stable release bumped the nix version again, if you push to the branch, it should re-trigger CI to run. Which will probably pass now.

@jonringer thx. I just rebased and force pushed the branch. however I see:

>  1 workflow awaiting approval
First-time contributors need a maintainer to approve running workflows. 

so perhaps the check will not run.

blitz:
> 21.05 is EOL, may just ask people to pin to a commit before this change if they want to be compatible with nix 2.3

It's not that easy. Nix 2.3 is for some reason still the default on NixOS 21.11. As long as Nix 2.3 is actively shipped, it's a bit user hostile to require 2.4 by default. We could check `builtins.nixVersion` and at least politely decline to work when `allRefs` is set?

Patryk27:
Hi! Thanks for this merge request - for the time being, I'm going to go with https://github.com/nix-community/naersk/pull/167, which allows for users to configure whether they want `allRefs` or not explicitly, and I'll revisit this topic once Nix 2.4 becomes more widely shipped. Thank y'all!

mhuesch:
@Patryk27 good call, thx and yw!

PR #213, opened 2021-12-08
Better error missing cargo lock

Anderssorby:
This will `throw` a more explanatory error message when Cargo.lock is missing form root.

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -356,7 +356,7 @@ let
       if builtins.pathExists cargolock-file then
         readTOML (cargolock-file)
       else 
-        throw "Naersk requires Cargo.lock to be available in root. Check that it is not in .gitignore when using git to filer (which flakes does)";
+        throw "Naersk requires Cargo.lock to be available in root. Check that it is not in .gitignore and stage it when using git to filer (which flakes does)";
 
     packageName =
       if ! isNull attrs.name
--- a/config.nix
+++ b/config.nix
@@ -356,7 +356,7 @@ let
       if builtins.pathExists cargolock-file then
         readTOML (cargolock-file)
       else 
-        throw "Naersk requires Cargo.lock to be available in root. Check that it is not in .gitignore and stage it when using git to filer (which flakes does)";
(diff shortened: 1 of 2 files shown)


config.nix
```diff
@@ -349,7 +349,14 @@ let
...
     # The cargo lock
-    cargolock = readTOML (root + "/Cargo.lock");
+    cargolock = 
+      let
+        cargolock-file = root + "/Cargo.lock";
+      in
+      if builtins.pathExists cargolock-file then
+        readTOML (cargolock-file)
+      else 
+        throw "Naersk requires Cargo.lock to be available in root. Check that it is not in .gitignore and stage it when using git to filer (which flakes does)";
```
blitz, line 359: nit:

```suggestion
        throw "Naersk requires Cargo.lock to be available in root. Check that it is not in .gitignore and stage it when using git to filter (which flakes does)";
```

I'm not sure what using "stage it when using to filter" means.

blitz (approved):
Definitely an improvement. Feel free to fix the typo and merge when the tests succeed.

config.nix
Anderssorby: stage it with `git add Cargo.lock` since flake filters the source using gitignore

Anderssorby:
@nmattia Awaiting your review.

config.nix
```diff
@@ -349,7 +349,14 @@ let
     toplevelCargotoml = readTOML (root + "/Cargo.toml");
 
     # The cargo lock
-    cargolock = readTOML (root + "/Cargo.lock");
+    cargolock = 
+      let
+        cargolock-file = root + "/Cargo.lock";
+      in
+      if builtins.pathExists cargolock-file then
```
nmattia, line 356: Did you try this with IFD, i.e. when the source is a built derivation in the store?

nmattia (approved):
This is a great idea!

config.nix
Anderssorby: Not if it is not in the CI.

PR #214, opened 2021-12-08
Handle custom registries

### LINKED ISSUE #186: Private crate registries
Hey!
I am starting to play around with nix to build Rust projects and I am experimenting on a crate that pulls in some dependencies from a private crate registry.
This leads to a failure when doing nix flake check - `error: cannot download download-XXX-0.3.0-beta.1 from any mirror.`
I can see that it's trying to pull the crate from https://crates.io, although a registry field is specified for that dependency in Cargo.toml.

Are private crate registries supported? If so, where can I find some documentation on how to fix my setup?

Anderssorby:
To fix #186 

* Load cargo config and control url in unpackCrate

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -6,11 +6,40 @@ let
 in
 rec
 {
+  # we differentiate 'src' and 'root'. 'src' is used as source for the build;
+  # 'root' is used to find files like 'Cargo.toml'. As often as possible 'root'
+  # should be a "path" to avoid reading values from the nix-store.
+  # Below we try to come up with some good values for src and root if they're
+  # not defined.
+  resolveSource = attrs:
+    let
+      hasRoot = ! isNull (attrs.root or null);
+      hasSrc = ! isNull (attrs.src or null);
+      isPath = x: builtins.typeOf x == "path";
+      root = attrs.root;
+      src = attrs.src;
+    in
+      # src: yes, root: no
+      if hasSrc && ! hasRoot then
+        if isPath src then
+          { src = lib.cleanSource src; root = src; }
+        else { inherit src; root = src; }
+        # src: yes, root: yes
+      else if hasRoot && hasSrc then
+        { inherit src root; }
+        # src: no, root: yes
+      else if hasRoot && ! hasSrc then
+        if isPath root then
+          { inherit root; src = lib.cleanSource root; }
+        else
+          { inherit root; src = root; }
+        # src: no, root: yes
+      else throw "please specify either 'src' or 'root'";
   # The list of _all_ crates (incl. transitive dependencies) with name,
   # version and sha256 of the crate
   # Example:
   #   [ { name = "wabt", version = "2.0.6", sha256 = "..." } ]
-  mkVersions = cargolock:
+  mkVersions = { cargolock, registries ? {} }:
     if builtins.hasAttr "metadata" cargolock then
 
       # TODO: this should nub by <pkg-name>-<pkg-version>
@@ -37,7 +66,10 @@ rec
           {
             inherit (p) name version;
             sha256 = p.checksum;
-          }
+          } // (lib.optionalAttrs (builtins.hasAttr "registry" p) {
+              inherit (registries.${p.registry}) index;
+            }
+          )
       ) (builtins.filter (builtins.hasAttr "checksum") cargolock.package)
     else [];
 
@@ -160,14 +192,14 @@ rec
 
   # A very minimal 'src' which makes cargo happy nonetheless
   dummySrc =
-    { cargoconfig   # string
+    { cargoConfigText   # string
     , cargotomls   # attrset
     , cargolock   # attrset
     , copySources # list of paths that should be copied to the output
     , copySourcesFrom # path from which to copy ${copySources}
     }:
       let
-        config = writeText "config" cargoconfig;
+        config = writeText "config" cargoConfigText;
         cargolock' = builtinz.writeTOML "Cargo.lock" cargolock;
         fixupCargoToml = cargotoml:
           let
@@ -193,7 +225,7 @@ rec
           { inherit copySources copySourcesFrom cargotomlss; }
           ''
             mkdir -p $out/.cargo
-            ${lib.optionalString (! isNull cargoconfig) "cp ${config} $out/.cargo/config"}
+            ${lib.optionalString (! isNull cargoConfigText) "cp ${config} $out/.cargo/config"}
             cp ${cargolock'} $out/Cargo.lock
 
             for tuple in $cargotomlss; do
--- a/build.nix
+++ b/build.nix
@@ -135,7 +135,7 @@ let
 
   nixSourcesDir = symlinkJoinPassViaFile {
     name = "crates-io";
-    paths = map (v: unpackCrate v.name v.version v.sha256)
+    paths = map (v: unpackCrate { inherit (v) name version sha256; })
       crateDependencies ++ [ unpackedGitDependencies ];
   };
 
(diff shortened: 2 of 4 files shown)


blitz:
@Anderssorby Looks pretty uncontroversial. Is there something in the test suite that tests this?

nmattia (commented):
LGTM, but I'm a bit rusty on how everything here works, so letting @blitz approve officially!

lib.nix
```diff
@@ -37,7 +66,10 @@ rec
           {
             inherit (p) name version;
             sha256 = p.checksum;
-          }
+          } // (if builtins.hasAttr "registry" p
+            then { inherit (registries.${p.registry}) index; }
+            else { }
+          )
```
jonringer, line 72:
```suggestion
          } // (lib.optionalAttrs (builtins.hasAttr "registry" p) {
              inherit (registries.${p.registry}) index;
            }
          )
```

otherwise LGTM

jonringer (dismissed):

jonringer:
I would say if the old workflow is still correct, then we should merge. At a glance, I don't see anything wrong. If something is wrong, then someone else can create an issue, and provide feedback as to how it's now working in their use case.

@Anderssorby do you mind squashing the fixup commit?

jonringer (approved):

Anderssorby:
So should we merge this?

blitz (changes requested):
> @blitz So should we merge this?

The code looks good, but one of the "slow" tests (as opposed to fast tests we execute in the CI) are still failing:

```
$ nix-build test.nix --show-trace -A rustlings 
error: please specify either 'src' or 'root'

       … while evaluating 'resolveSource'

       at /home/julian/src/own/naersk/lib.nix:14:19:

           13|   # not defined.
           14|   resolveSource = attrs:
             |                   ^
           15|     let

       … from call site

       at /home/julian/src/own/naersk/default.nix:28:16:

           27|       # Config for cargo itself
           28|       source = libb.resolveSource arg;
             |                ^
           29|       configFile = source.root + "/.cargo/config.toml";

       … while evaluating 'optionalString'
```

<details>
```

       at /nix/store/vqd0zjq6qjs5zz2m1nnzp4w5glf7ga7r-nixpkgs-src/lib/strings.nix:202:5:

          201|     # String to return if condition is true
          202|     string: if cond then string else "";
             |     ^
          203|

       … from call site

       at /home/julian/src/own/naersk/lib.nix:228:15:

          227|             mkdir -p $out/.cargo
          228|             ${lib.optionalString (! isNull cargoConfigText) "cp ${config} $out/.cargo/config"}
             |               ^
          229|             cp ${cargolock'} $out/Cargo.lock

       … while evaluating the attribute 'buildCommand' of the derivation 'dummy-src'

       at /nix/store/vqd0zjq6qjs5zz2m1nnzp4w5glf7ga7r-nixpkgs-src/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'src' of the derivation 'rustlings-deps-1.3.0'

       at /nix/store/vqd0zjq6qjs5zz2m1nnzp4w5glf7ga7r-nixpkgs-src/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'builtDependencies' of the derivation 'rustlings-1.3.0'

       at /nix/store/vqd0zjq6qjs5zz2m1nnzp4w5glf7ga7r-nixpkgs-src/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

```
</details>

Patryk27:
Hi, @Anderssorby - thanks for this merge request! Would you mind rebasing it & addressing the comment from @blitz?

Anderssorby:
Sure, I can take a look later.

SkamDart:
@Anderssorby Awesome work. I would also like to see this feature get merged in. Is there anything I can do to help you close out this PR?

Frarese:
Hello, I would also like to see this feature merged and I tried to understand why the test wasn't passing.
Apparently we check for the attributes `root` or `src` but in the case of `rustling` we pass the source like this:

```
rustlings = naersk.buildPackage sources.rustlings;
```

instead of this:

```
rustlings = naersk.buildPackage {
    src = sources.rustlings;
};
```

There are two ways to fix this issue:

1. We keep the code as is and change the test to reflect that we always need to specify `src` or `root`;
2. Or, [as i did in this commit](https://github.com/nix-community/naersk/commit/f1ef58512b51a99c8d43fd805cfc8b71b77e01a3), if both attributes are missing we pass `attrs` itself as a source.

Let me know if I can contribute further :relaxed:

Patryk27:
> Let me know if I can contribute further

Thanks for analyzing the issue, your second approach seems the most fitting - I'll try to rebase this pull request with your patch applied over the weekend 🙂

Patryk27:
Hmm, maybe I'm misunderstanding something, but the solution presented in this pull request can't work, can it? 👀 

- `host`, `index` and `url` as passed to `unpackCrate` are never really set anywhere (besides being assigned their default values),
- We don't checkout the registry's index, which we need to do in order to load the registry's configuration (https://doc.rust-lang.org/cargo/reference/registries.html#index-format) - otherwise naersk can't possibly know where to download the package from.

So, say, we're given a `Cargo.lock` that says:

```toml
[[package]]
name = "dep"
version = "0.1.0"
source = "registry+https://git.some-company.com/rust/crates-index"
checksum = "bae3d8de1b7fd1fef6c2da3130a7d06d32499fd5292a9c1309681ac79e98c643"
```

To actually download that package naresk would have to:

1. Checkout `https://git.some-company.com/rust/crates-index`,
2. Load `config.json` from in there, parse its `dl` key,
3. Call `unpackCrate` with the results of that.

... and I think the first step is fundamentally problematic in Nix, since we don't know at which commit we should checkout that hypothetical `rust/crates-index` (considering that `checksum` in lockfile refers to the package itself, not its registry's index).

I'd say that in order to have some minimum viable product we should extend `buildPackage`, so that instead of parsing `.cargo/config.toml` it simply accepts a new parameter describing registries required to build the package:

```nix
{ ... }:

naersk.buildPackage {
  src = ./.;
  
  registries = {
    # crates-io remains implied, ofc.
    
    "custom" = {
      dl = "https://my-registry.io/api/v1/crates/{version}/{crate}/whatever/{lowerprefix}";
    };
  };
}
```

I think this would solve all of the problems here, but I'm leaving this one up to discussion :smile:

Frarese:
So I started looking into it again and have an implementaton of `buildPackge` that looks like this:

```nix
buildPackage = args@{registries? {}, ... }:
```

that passes the fast tests but fails the slow ones:

```
Building
NOT testing on darwin, enabling sandbox
error: value is a string with context while a set was expected

       at /home/frarese/repos/naersk/test/slow/talent-plan/default.nix:12:19:

           11|
           12|   talent-plan-3 = naersk.buildPackage "${sources.talent-plan}/rust/projects/project-3";
             |                   ^
           13| }

```

should I just make a `registriesCustomDl` attribute so that we can maintain the signature for the function or is there a better way to include `registries` in the options for `buildPackage` ?

Patryk27:
To create a new parameter, you should just add it to `build.nix` - that will not only later include it in README.md, but also compose neatly with rest of the library 🙂

secana:
Hi! What is missing to get this PR through? I'm the maintainer of https://kellnr.io and I would really like to be able to use a private registry for Rust with Nix. Unfortunately, I'm very new to Nix, so I lack the skill to implement the feature myself.

nmattia:
> I'd suggest switching to [Crane](https://github.com/ipetkov/crane), though - it's (imo) better designed and it's kept very much up to date in terms of support and features.

Oh wow, Crane looks nice! Since it's maintained, well documented and supports many more features I think it's best to archive naersk and point people to Crane; best if people don't try setting up naersk just to realize it's deprecated anyway.

Unless people complain I'll do that soon!

PR #215, opened 2021-12-16
ci: bump nixpkgs versions

ipetkov:
* Update to a more recent revision of nixpkgs
* Add nixpkgs-21.11 for testing
* Add nixpkgs-21.05 for testing
* Drop nixpkgs-20.09, branch is EOL
* Drop nixpkgs-20.05, branch is EOL
* Bump ripgrep-all to v0.9.6 to make tests pass

### DIFF
--- a/nix/sources.nix
+++ b/nix/sources.nix
@@ -98,7 +98,10 @@ let
       saneName = stringAsChars (c: if isNull (builtins.match "[a-zA-Z0-9]" c) then "_" else c) name;
       ersatz = builtins.getEnv "NIV_OVERRIDE_${saneName}";
     in
-      if ersatz == "" then drv else ersatz;
+      if ersatz == "" then drv else
+        # this turns the string into an actual Nix path (for both absolute and
+        # relative paths)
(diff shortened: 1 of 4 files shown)


nix/sources.nix
```diff
@@ -98,7 +98,10 @@ let
       saneName = stringAsChars (c: if isNull (builtins.match "[a-zA-Z0-9]" c) then "_" else c) name;
       ersatz = builtins.getEnv "NIV_OVERRIDE_${saneName}";
     in
-      if ersatz == "" then drv else ersatz;
+      if ersatz == "" then drv else
```
nmattia, line 101: What was the issue here? Would you like to create a ticket in https://github.com/nmattia/niv?

nmattia (approved):
Thanks a lot!

nix/sources.nix
ipetkov: Ah I should have mentioned... This was volunteered by niv itself (IIRC bumping the source version from 23 to 24).

Diff comes from invoking `niv init` and I decided to keep it for the sake of being on the latest-and-greatest version of everything :)

nmattia: hah, I see! Thanks

PR #219, opened 2021-12-19
Correctly fetch git tags

Ma27:
Without this change you'd get errors like this when adding a
`git`-dependency to a `Cargo.toml` which is pinned to a tag:

    fetching Git repository 'git://github.com/ma27/nixpkgs-fmt'fatal: couldn't find remote ref refs/heads/v1.2.0-rnix-0.10
    error: program 'git' failed with exit code 128

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -151,7 +151,7 @@ rec
                   } // lib.optionalAttrs (val ? branch) {
                     ref = val.branch;
                   } // lib.optionalAttrs (val ? tag) {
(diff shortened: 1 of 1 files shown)


nmattia (approved):
LGTM!

Anderssorby (approved):
LGTM too

blitz (approved):
Thanks @Ma27! Merging this, because it creates user pain, but it would be great if we had a test for this as well.

PR #220, opened 2021-12-23
build: skip invalid manifests

FlorianFranzen:
This patch makes sure that only crates that pass some rudimentary check are unpacked in `git-deps`. This avoids git repos with random and/or broken `Cargo.toml` files to break your whole build. 

The proper fix would be to unpack needed dependencies, but that is a bit more involved, so this should offer a quick and simple fix for now.

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -90,7 +90,7 @@ let
   #     * grab whatever is surrounded with `"`s.
   #   The last step is very, very slow.
   unpackedGitDependencies = runCommand "git-deps"
-    { nativeBuildInputs = [ jq ]; }
+    { nativeBuildInputs = [ cargo jq ]; }
     ''
       log() {
         >&2 echo "[naersk]" "$@"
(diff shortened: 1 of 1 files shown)


jonringer:
Wouldn't it be preferable to just fix the toml file?

FlorianFranzen:
@jonringer I patched this even more by only installing the dependencies request (see [here](https://github.com/nix-community/naersk/pull/167#issuecomment-1000318797)) in which case this would be no longer needed, but I thought this is a good quick fix.

Patryk27:
Hi, thanks for this merge request; as far as I understand, https://github.com/nix-community/naersk/pull/167 already solves it, right? (unless there's some additional depth to `cargo verify-project` that I don't see 🙂)

So in the effort of cleaning-up issues in this repository, I'll just go ahead and close this merge request - but if you feel we could use `cargo verify-project`, please feel free to re-open it 🙂 

Thanks again!

PR #227, opened 2022-03-07
Support different `lib.crate-type`

NickHu:
This is particularly important for the case of WASM binaries, which are necessarily compiled with `crate-type=cdylib`. Without this, when compiling for `CARGO_BUILD_TARGET=wasm32-unknown-unknown`, the `-deps` derivation strips the `[lib]` section of the `Cargo.toml`, so only `lib*.d` and `lib*.rlib` files artifacts are generated under `target/`. Then when the main derivation is compiled, the `[lib]` section is not stripped, so it must recompile all the dependencies again for `cdylib` in order to generate a wasm binary.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -174,7 +174,7 @@ rec
             attrs =
               # Since we pretend everything is a lib, we remove any mentions
               # of binaries
-              removeAttrs cargotoml [ "bin" "example" "lib" "test" "bench" "default-run" ]
+              removeAttrs cargotoml [ "bin" "example" "test" "bench" "default-run" ]
                 // lib.optionalAttrs (builtins.hasAttr "package" cargotoml) ({ package = removeAttrs cargotoml.package [ "default-run" ] ; })
                 ;
           in
@@ -209,7 +209,11 @@ rec
                 pushd $out/$member > /dev/null
                 mkdir -p src
                 # Avoid accidentally pulling `std` for no-std crates.
-                echo '#![no_std]' >src/lib.rs
+                cat <<EOF >src/lib.rs
+                #![no_std]
+                #[panic_handler]
+                fn panic(_info: &core::panic::PanicInfo) -> ! { loop {} }
+            EOF
                 # pretend there's a `build.rs`, otherwise cargo doesn't build
                 # the `[build-dependencies]`. Custom locations of build scripts
                 # aren't an issue because we strip the `build` field in


lib.nix
```diff
@@ -209,7 +209,11 @@ rec
                 pushd $out/$member > /dev/null
                 mkdir -p src
                 # Avoid accidentally pulling `std` for no-std crates.
-                echo '#![no_std]' >src/lib.rs
+                cat <<EOF >src/lib.rs
+                #![no_std]
+                #[panic_handler]
+                fn panic(_info: &core::panic::PanicInfo) -> ! { loop {} }
```
nmattia, line 215: what are these two lines for?

lib.nix
```diff
@@ -174,7 +174,7 @@ rec
             attrs =
               # Since we pretend everything is a lib, we remove any mentions
               # of binaries
-              removeAttrs cargotoml [ "bin" "example" "lib" "test" "bench" "default-run" ]
+              removeAttrs cargotoml [ "bin" "example" "test" "bench" "default-run" ]
```
nmattia, line 177: what happens if the project mentions a `[lib]` that isn't `src/lib.rs`? Won't cargo try to build that and, not finding any files, fail?

nmattia (commented):
Few questions!

lib.nix
NickHu: If you compile with `crate-type=cdylib`, and also use `no_std`, the compilation will fail because no panic handler has been provided (the bit that unwinds the stack trace, usually provided by the standard library). This is just a dummy stub to get compilation to succeed. See https://doc.rust-lang.org/nomicon/panic-handler.html for more details.

lib.nix
NickHu: To be honest, this change in particular is the one I'm not sure about. In order for WASM dependencies to be built, `lib.crate-type` needs to be retained. This was merely the quickest fix I could think of, I didn't really consider the other effects, and perhaps there's a better way to achieve the same goal.

lib.nix
```diff
@@ -209,7 +209,11 @@ rec
                 pushd $out/$member > /dev/null
                 mkdir -p src
                 # Avoid accidentally pulling `std` for no-std crates.
-                echo '#![no_std]' >src/lib.rs
+                cat <<EOF >src/lib.rs
+                #![no_std]
```
nmattia, line 213:
```suggestion
                #![no_std]
                # If you compile with crate-type=cdylib, and also use no_std, 
                # the compilation will fail because no panic handler has been
                # provided (the bit that unwinds the stack trace, usually provided 
                # by the standard library). This is just a dummy stub to get
                # compilation to succeed. 
                # See https://doc.rust-lang.org/nomicon/panic-handler.html for more details.
```

lib.nix
nmattia: Thanks! Added a code suggestion; worth having the clarification near the code.

nmattia (commented):
Thanks!

lib.nix
NickHu: Any example which uses `wasm-bindgen` for instance:
https://github.com/rustwasm/wasm-bindgen/tree/main/examples

njaremko:
@NickHu Are you still working on this? If not, I'll pick it up because I need this working 😆

PR #232, opened 2022-04-04
Support dependencies with slashes in their names

Patryk27:
When a dependency gets fetched from Git, it's possible for its name to
contain slashes - e.g. because the dependency's branch name contains
one:

```
something-something = { git = "...", branch = "JIRA-123/dev" }
```

Currently, because we don't sanitize those slashes, trying to compile
such `Cargo.toml` / `Cargo.lock` would fail - e.g. given a dependency
from branch `JIRA-123/dev`, we'd try to extract it into path
`JIRA-123/dev` instead of something flat such as `JIRA-123_dev`.

This commit fixes this issue by sanitizing all the `nkey`s - basically,
for all dependency paths, we replace `/` with `_`.

The choice of underscore is arbitrary, just seemed like a good fit.


build.nix
```diff
@@ -335,24 +345,21 @@ let
       runHook postConfigure
     '';
 
-    buildPhase =
```
Patryk27, line 338: This was not changed, only reformatted (one nesting level less, as per rest of the scripts in this file).

test.nix
```diff
@@ -174,9 +174,37 @@ let
       git-dep = naersk.buildPackage {
         doCheck = true;
         src = ./test/git-dep;
+      };
+
+      git-dep-by-branch = naersk.buildPackage {
```
Patryk27, line 179: This test was not present previously in here (even though the directory already existed); not sure why, so I've brought it back.

nmattia (commented):
Thanks! I haven't worked on the codebase in a while so maybe those questions are silly:

* What's the problem with actually unpacking branches as nested dirs? I.e. I'm guessing we could do something like `mkdir -p $out/$nkey`, which would lead to `$out/foo/bar` right?
* I don't remember what we unpack exactly, but what happens if we have say branches `foo/bar` and `foo_bar`? Won't the rewrite (`s$/$_/`) for `foo/bar` overwrite `foo_bar`?

Thanks a ton for the tests!

Patryk27:
> What's the problem with actually unpacking branches as nested dirs? 

Cargo expects a flat structure (i.e. all extracted dependencies must be at the same "directory level"), and so unpacking into nested directories makes those packages invisible to Cargo.

> but what happens if we have say branches [...]

`nkey` contains package's checksum (e.g. `"rand-0.7.0-pre.1-703452450770d4b2bb0b117dfc83aea5ba61ec60"`), so the chance of a collision happening due to this sanitization is practically zero 🙂 -- in theory, when a collision happens, we'll hit:
```rust
if [ -d "$dest" ]; then
  log "Crate was already unpacked at $dest"
else
```
... and presumably the compilation will then fail due to a missing dependency.

But since `nkey` has those checksum, I think there's no need to worry about this case.

nmattia (dismissed):
Thanks a lot, looking good!

Patryk27:
tl;dr I've pushed a new commit with some additional fixes

## First issue

A it turns out, the CI failure was caused by a mismatch in Git versions - the repository created by me (and stored into `test/.../dep/git`) was built with Git 2.35.1, while the pipelines run on Git 2.34.0 or so, making `builtins.fetchGit` go haywire.

To avoid this issue (and prevent it from happening in the future), I've slightly changed the test, so that instead of having binary Git repository inside the test's directory, we generate the repository dynamically before the test:

```nix
{
  dep = pkgs.runCommand "dep" {  
    buildInputs = [ pkgs.git ];
  } ''
    mkdir $out
    cd $out
    cp -ar ${./test/git-dep-by-branch-with-slash/dep}/* .

    git init --initial-branch=with/slash
    git add .
    git config user.email 'someone'
    git config user.name 'someone'
    git commit -am 'Initial commit'
  '';
};
```

## Second issue

On Nix 2.7, `nix-build test.nix` failed to evaluate our new test, saying:
```
the string '< Cargo.toml's contents >' is not allowed to refer to a store path
```

I've narrowed down the issue to the fact that this test calls `naersk.buildPackage` with a _dynamically-built_ source:
```nix
let
  dep = pkgs.runCommand "dep" {
    # ...
  } ''
    # ...
  '';

  app = pkgs.runCommand "app" {
    # ...
  } ''
    # this string has `${dep}` somewhere inside of it, so `app` depends on `dep`
  '';

in
naersk.buildPackage {
  src = app;
  # ...
};
```
... and that makes `builtins.fromTOML` angry, because it expects a context-less string.

To avoid this issue, I've used `builtins.unsafeDiscardStringContext` and dropped a comment explaining the rationale.

blitz (approved):
Thanks for the fix and also for the test and pointers to information about the string contexts!

From my end everything looks fine. If some Nix guru would look at the string context issue, that would be good though.

blitz:
I'll give @Anderssorby a bit of time to look at this. Will merge in a couple of days, if no issue pops up.

Issue #230, opened 2022-04-01
Symlinks are not resolved after a submodule is cloned


sekunho:
Heyah. Thanks for the awesome work!

I have a cargo dependency that contains a submodule, and it doesn't seem like it gets cloned properly by (I think).

Dep's submodule: https://github.com/lambda-fairy/maud/blob/main/maud_macros/src/escape.rs
nix version: 2.5.1
`flake.nix`:

```nix
{
  description = "";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs";
    flake-utils.url = "github:numtide/flake-utils";
    naersk.url = "github:nix-community/naersk";
  };

  outputs = { self, nixpkgs, nixos-unstable, flake-utils, naersk }:
    flake-utils.lib.eachSystem [ "x86_64-linux" ] (system:
      let pkgs = nixpkgs.legacyPackages.${system};
           naersk-lib = naersk.lib."${system}".override {
             cargo = pkgs.cargo;
             rustc = pkgs.rustc;
           };
      in
      rec {
        packages.foo = naersk-lib.buildPackage {
          pname = "foo";
          root = ./.;
        };

        defaultPackage = packages.foo;

        apps.foo = flake-utils.lib.mkApp {
          drv = packages.foo;
        };

        defaultApp = apps.foo;

        devShell = pkgs.mkShell {
          nativeBuildInputs = with pkgs; [
            pkgs.rustc
            pkgs.cargo
          ];
        };
      });
}
```

`nix build` log:

```
Compiling maud_macros v0.23.0 (https://github.com/lambda-fairy/maud?branch=main#e6787cd6)
error: could not compile `maud_macros` due to 3 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed
error[E0583]: file not found for module `escape`
 --> /sources/maud_macros-0.23.0-e6787cd62165a075c7f16a32f8bbacc398f52d13/src/lib.rs:9:1
  |
9 | mod escape;
  | ^^^^^^^^^^^
  |
  = help: to create the module `escape`, create file "/nix/store/shixfmbv64zzi9i0rzb9c8870ydm7cf2-crates-io/maud_macr>


error[E0425]: cannot find function `escape_to_string` in module `escape`
   --> /sources/maud_macros-0.23.0-e6787cd62165a075c7f16a32f8bbacc398f52d13/src/generate.rs:272:17
    |
272 |         escape::escape_to_string(string, &mut self.tail);
[...]

Issue #158, opened 2021-03-16
preferLocal for download/unpack derivations


ghost:
I have some Nix remote builders configured on my system globally. Every time I try to build something that's packaged wtih naersk, it will copy the dependency archives to remote builders, extract them there and download the unpacked sources again. This is much slower than building the `unpack` derivations locally.

PR #234, opened 2022-04-19
Improve unpacking dependencies

Patryk27:
Closes https://github.com/nix-community/naersk/issues/230.
Closes https://github.com/nix-community/naersk/issues/158.

# Abstract

## Symlinks

This pull request takes a bit different approach to solve #230 - instead of going with:
```
cp -rL $(dirname $toml) $dest
```
... which has this downside of "if the symlink is referred to from multiple places, we'll copy the symlink multiple times, wasting space", I've decided to go with:
```
ln -s $(dirname $toml)/* $dest
```

This approach not only solves the original symlink-issue, but it also reduces the overall space required for processing Git dependencies - instead of having each dependency duplicated (once through `fetchGit` and then through `cp` inside `runCommand`), we're downloading the dependency just once and then keep referring to it through symlinks.

I've also changed the way Git dependencies are unpacked, so that instead of having one giant `git-deps` derivation with all Git dependencies build through it, we do `(map unpackGitDependency gitDependencies)` -- this should make caching those dependencies better, since now if only one of `n` Git dependencies changes, we'll only have to rediscover `Cargo.toml` of that particular dependency instead of going through all of them again.

## runCommandLocal

I've switched a few places from using `runCommand` to `runCommandLocal` - a brief `nix build -vv` with a remote builder tells me it reduces the amount of network-hopping, improving the speed a bit.


rust/1_38_0.nix
```diff
@@ -1,22 +0,0 @@
-import ./default.nix {
```
Patryk27, line 1: The only place were those files were referred to was through there:
```
// pkgs.lib.optionalAttrs (nixpkgs == "nixpkgs-20.03" && pkgs.stdenv.isLinux) muslTests
```
... but there's no mention of `nixpkgs-20.03` anywhere else, so those `rust/*` files (required for some of the `muslTests`) were effectively not used anyway.

Instead of rewriting them to use a newer nixpkgs, I've decided that it's probably best if we get rid of them and later create a separate wasm/musl/etc. testsuite that doesn't require compiling rustc from scratch (e.g. by using nixpkgs-mozilla).

test.nix
```diff
@@ -1,317 +1,5 @@
...
-              echo target directory is empty: "$dep"
-              return 1
-            fi
-          done
-
-          # Success
-          touch $out
-        '';
-
-      simple-dep = naersk.buildPackage {
```
Patryk27, line 140: To avoid ending up with one, large Nix file, those tests were refactored into `test/fast/test-name/default.nix`

rust/1_38_0.nix
nmattia: Ah, I see! Then this was definitely overlooked when we upgraded from `nixpkgs-20.03`. I think there was one particular use case for those tests at $WORK but I don't mind if we simply get rid of them. Either way let's not add a dependency on nixpkgs-mozilla.

nmattia (commented):
Thanks for the contribution! This is a pretty big PR, which makes it a little hard to review. Is there any chance we could split it? I'm struggling to wrap my head around what actually changed regarding the git unpack logic, so a PR focused on that would be great!

But I'm sure there's someone out there with more brain cycles available, so I'm happy to just let them review!

nmattia:
Sounds awesome, thanks @Patryk27 ! Let's keep this one open in case you think the split isn't worth the time :)

Patryk27:
Okie, rebased 🙂 Lemme know if that's reviewable now or still needs some refinements!

nmattia:
I've created a branch with some changes to `build.nix` which I think would make this PR much easier to review:

https://github.com/nix-community/naersk/compare/improve-dependency-handling...nm-improve-dependency-handling

They all appeared to be cosmetic changes (code was reformatted), some function renaming (`symlinkJoin`) and some comment/documentation removal (which I didn't quite understand).

Did I miss something, or can those changes really be reverted? If they can be reverted, can you do something similar with the other files modified by this PR (#234) so that we can focus the review on what changed? Thanks a lot!

nmattia:
I've created a branch with some changes to `build.nix` which I think would make this PR much easier to review:

https://github.com/nix-community/naersk/compare/improve-dependency-handling...nm-improve-dependency-handling

They all appeared to be cosmetic changes (code was reformatted), some function renaming (`symlinkJoin`) and some comment/documentation removal (which I didn't quite understand).

Did I miss something, or can those changes really be reverted? If they can be reverted, can you do something similar with the other files modified by this PR (#234) so that we can focus the review on what changed? Thanks a lot!

Patryk27:
Sure, I've reverted the changes - let's focus on the fixes first and I'll create a separate MR for the stylistic changes after we're done with this one 🙂

nmattia (approved):
@Patryk27 my bad, sorry for dropping the ball on this!

Looks good, I haven't tested it but as long as the tests pass and incrementality is still working, it's good enough for me :)

Thanks a lot for this!

Issue #237, opened 2022-05-14
Wrapping programs is hard


scvalex:
I've got a Rust app that needs LD_LIBRARY_PATH to run.

I think the usual Nix way to support this is to `wrapProgram` the executable, so I tried adding a `postInstall` hook to `naersk-lib.buildPackage` in my flake:

```nix
        packages.sixty-two = naersk-lib.buildPackage {
          src = ./.;
          doCheck = true;
          pname = "sixty-two";
          nativeBuildInputs = [ pkgs.makeWrapper ];
          buildInputs = with pkgs; [
            xorg.libxcb
          ];
          postInstall = ''
            wrapProgram "$out/bin/sixty-two" --prefix LD_LIBRARY_PATH : "${libPath}"
          '';
        };
```

Running `nix build` fails with `Cannot wrap '/nix/store/68rf1hi47h6f7kchgja73kai1d1f0r74-sixty-two-deps-0.1.0/bin/sixty-two' because it is not an executable file`.  Note the "deps" in the store path.

```
$ nix build
warning: Git tree '/home/scvalex/repo/sixty-two' is dirty
error: builder for '/nix/store/q4caqhx4nlyxf76kfwxz64vk195zqzs6-sixty-two-deps-0.1.0.drv' failed with exit code 1;
       last 10 log lines:
       > Builder called die: Cannot wrap '/nix/store/68rf1hi47h6f7kchgja73kai1d1f0r74-sixty-two-deps-0.1.0/bin/sixty-two' because it is not an executable file
       > Backtrace:
       > 7 assertExecutable /nix/store/wdk5bvld0w00ca4dlpjmbqw2mjx3ppgm-hook/nix-support/setup-hook
       > 136 wrapProgram /nix/store/wdk5bvld0w00ca4dlpjmbqw2mjx3ppgm-hook/nix-support/setup-hook
       > 84 _callImplicitHook /nix/store/ydi7m5hgm1qd4r67z0vp8vjr14s8jh3p-stdenv-linux/setup
       > 101 _eval /nix/store/ydi7m5hgm1qd4r67z0vp8vjr14s8jh3p-stdenv-linux/setup
       > 44 runHook /nix/store/ydi7m5hgm1qd4r67z0vp8vjr14s8jh3p-stdenv-linux/setup
       > 1361 genericBuild /nix/store/ydi7m5hgm1qd4r67z0vp8vjr14s8jh3p-stdenv-linux/setup
[...]

PR #238, opened 2022-05-15
Add support for the `postInstall` hook

Patryk27:
During compilation, we build two derivations - one for the dependencies
and another for the user's crate itself. Passing `postInstall` to
`buildPackage` pass-throughs it into both derivations, which is not what
most users would probably expect.

This commit adds `postInstall` as an explicit configuration option (so
that it's neatly documented), zeroing it out for the dependencies-only
derivation.

I don't think there's any use case for something like `postInstallDeps`,
so it's not introduced here.

Closes https://github.com/nix-community/naersk/issues/237


nmattia (approved):
LGTM!

In the future, would be great if you could submit unrelated (cosmetic + README) changes separately! :)

Patryk27:
> In the future, would be great if you could submit unrelated (cosmetic + README) changes separately! :)

Sure 🙂 fwiw, I've pushed them as separate commits, since GitHub allows to review stuff commit-by-commit which is basically like creating `n` pull requests, but easier -- but if you find separate pull request easier, then I can't see a reason why not 😄

PR #243, opened 2022-06-03
force fetchGit allRefs when a dependency is specified with "rev"

goertzenator:
The usage of git "rev" dependencies in Cargo implies that the rev
will be found regardless of ref.  This is currently not true in naersk;
the following dependency fails:

usbfs = {git = "https://github.com/goertzenator/usbfs-rs", rev = "7c35b46e41ad76f838cd99b493f9337f475b5b70"}

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -111,7 +111,7 @@ rec
           ref = lock.branch;
         } // lib.optionalAttrs (lock ? tag) {
           ref = lock.tag;
-        } // lib.optionalAttrs (gitAllRefs || lock ? rev) {
+        } // lib.optionalAttrs ((lib.versionAtLeast builtins.nixVersion "2.4") && (gitAllRefs || lock ? rev)) {
           allRefs = true;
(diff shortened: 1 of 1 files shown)


Patryk27:
Hi 🙂 

Thanks for this pull request; we've already considered doing that in the past (https://github.com/nix-community/naersk/pull/211), but eventually settled on a dedicated option, because `allRefs` was not supported by Nix 2.3 (which was still pretty prevalent at the time).

I see NixOS 22.05 ships with Nix 2.8 and https://nixos.org/download.html provides Nix 2.9, so I think it's time to revisit that decision - but, just in case, could you introduce a version check there?

Patryk27:
whoopsie, miss-click!

goertzenator:
Thanks for the response.  I will add the suggested change in the next few days.

goertzenator:
Additional reference information on this topic: https://github.com/NixOS/nix/issues/5128

Patryk27 (approved):
Looks neat, thanks!

Issue #274, opened 2023-02-20
crate that is necessary fo building is removed from the `crates` directory when trying to build `nu-shell` 


MangoIV:
trying to build `nu-shell` fails with 

```
nu-deps> error: failed to load manifest for workspace member `/build/dummy-src/crates/nu-cli`
nu-deps> Caused by:
nu-deps>   failed to load manifest for dependency `nu-color-config`
nu-deps> Caused by:
nu-deps>   failed to read `/build/dummy-src/crates/nu-color-config/Cargo.toml`
nu-deps> Caused by:
nu-deps>   No such file or directory (os error 2)
nu-deps> [naersk] cargo returned with exit code 101, exiting
```

Apparently this is because `naersk` does source cleaning, but removes things that are needed for the build. 

In this example, it removes `nu-color-config`

before cleaning: 

```
────┬────────────────────────────────┬──────┬─────────┬─────────────
  # │              name              │ type │  size   │  modified   
────┼────────────────────────────────┼──────┼─────────┼─────────────
  0 │ crates/README.md               │ file │   795 B │ an hour ago 
  1 │ crates/nu-cli                  │ dir  │ 4.0 KiB │ an hour ago 
  2 │ crates/nu-color-config         │ dir  │ 4.0 KiB │ an hour ago 
  3 │ crates/nu-command              │ dir  │ 4.0 KiB │ an hour ago 
  4 │ crates/nu-engine               │ dir  │ 4.0 KiB │ an hour ago 
  5 │ crates/nu-explore              │ dir  │ 4.0 KiB │ an hour ago 
  6 │ crates/nu-glob                 │ dir  │ 4.0 KiB │ an hour ago 
  7 │ crates/nu-json                 │ dir  │ 4.0 KiB │ an hour ago 
  8 │ crates/nu-parser               │ dir  │ 4.0 KiB │ an hour ago 
  9 │ crates/nu-path                 │ dir  │ 4.0 KiB │ an hour ago 
 10 │ crates/nu-plugin               │ dir  │ 4.0 KiB │ an hour ago 
 11 │ crates/nu-pretty-hex           │ dir  │ 4.0 KiB │ an hour ago 
 12 │ crates/nu-protocol             │ dir  │ 4.0 KiB │ an hour ago 
 13 │ crates/nu-system               │ dir  │ 4.0 KiB │ an hour ago 
 14 │ crates/nu-table                │ dir  │ 4.0 KiB │ an hour ago 
[...]

Issue #273, opened 2023-02-17
Naersk handles multi-toolchain builds incorrectly


6e6f61:
I'm trying to build a project (a native binary and WASM modules), but it seems Naersk is setting Cargo's build targets incorrectly.

**Example**: https://gist.github.com/6e6f61/90ae0560f64d7c6911107c99d1663bd0 (forgive the inability to create directories in Gists, I replaced `/`s with `_`).

If you build this example, you'll notice `result` contains `bin/a`, `lib/b.wasm` (as expected), but also `bin/a.wasm`, which wasn't requested.

When the `CARGO_BUILD_TARGET` directive is removed from `packages.b`, everything builds correctly.

As noted in the `flake.nix` example, adding `CARGO_BUILD_TARGET = "x86_64-unknown-linux-gnu"` to `packages.a`'s derivation still results in Cargo trying to build a WASM version too.

6e6f61:
I'll note treating the workspace's members as entirely separate crates resolves this, i.e.:

```
src = ./b;
# src = ./.;
# targets = [ "b" ];
```

But I'll leave this issue open as I can't imagine the prior behaviour is deliberate and I'd suppose the semantics of these two configurations differ somehow.

PR #276, opened 2023-02-25
Improve Cargo.toml discovery algorithm

Patryk27:
Our current Cargo.toml discovery algorithm is based on reading the main Cargo.toml and looking at its `workspace.members` - unfortunately, as it turns out there's actually no requirement for all workspace crates to be listed there, and some applications do fancy things like:

```toml
[workspace]
members = [ "crates/foo" ]

[dependencies]
foo = { path = "crates/foo" }
bar = { path = "crates/bar" } # whoopsie
```

... which doesn't play nice with Naersk's incremental builds (that is, the compilation fails unless `singleStep = true;` is turned on).

This commit changes the logic to use a different approach: now we simply recursively scan the root directory, noting down all the Cargo.tomls we find; Crane seems to use the same algorithm.

I think the only case where this approach falls short are spurious Cargo.tomls that are present somewhere in the source directory but are actually unused - for instance, imagine I'm writing a Cargo clone where I've got files like `src/tests/broken-manifest/Cargo.toml` - in cases like these, the current approach will cause the build to fail, because we will try to fixup a broken Cargo.toml, thinking it's used somewhere.

We could try to handle cases like these by using an even different approach: instead of traversing the source tree, we could load the main Cargo.toml and traverse its `workspace.members` + all path-based dependencies recursively, noting down all the **explicitly present** Cargo.tomls.

That is, given a top-level manifest of:

```toml
[workspace]
members = [ "crates/foo" ]

[dependencies]
foo = { path = "crates/foo" }
bar = { path = "crates/bar" }
```

... we'd note down and load `crates/foo/Cargo.toml` + `crates/bar/Cargo.toml`, and then load _their_ path-based dependencies, to find all transitive Cargo.tomls that belong to this workspace.

(that is, we have to scan the crates recursively, because `bar` might depend on `crates/zar` - and that's not an imaginary edge case, that's actually what happens inside Nushell, for instance.)

I don't have enough time to implement this more extensive approach now, and IMO the improvement presented here is still an improvement, so I think it's safe to go with the new-but-still-subpar approach here and consider the better algorithm in the future :-)

# Testing

All tests pass, I've also made sure that incremental builds are still incremental.

Closes https://github.com/nix-community/naersk/issues/274.
Closes https://github.com/nix-community/naersk/issues/273.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,6 @@ process, rest is passed-through into `mkDerivation`.
 | `override` | An override for all derivations involved in the build. Default: `(x: x)` |
 | `overrideMain` | An override for the top-level (last, main) derivation. If both `override` and `overrideMain` are specified, _both_ will be applied to the top-level derivation. Default: `(x: x)` |
 | `singleStep` | When true, no intermediary (dependency-only) build is run. Enabling `singleStep` greatly reduces the incrementality of the builds. Default: `false` |
-| `targets` | The targets to build if the `Cargo.toml` is a virtual manifest. |
 | `copyBins` | When true, the resulting binaries are copied to `$out/bin`. <br/> Note: this relies on cargo's `--message-format` argument, set in the default `cargoBuildOptions`. Default: `true` |
 | `copyLibs` | When true, the resulting binaries are copied to `$out/lib`. <br/> Note: this relies on cargo's `--message-format` argument, set in the default `cargoBuildOptions`. Default: `false` |
 | `copyBinsFilter` | A [`jq`](https://stedolan.github.io/jq) filter for selecting which build artifacts to release. This is run on cargo's [`--message-format`](https://doc.rust-lang.org/cargo/reference/external-tools.html#json-messages) JSON output. <br/> The value is written to the `cargo_bins_jq_filter` variable. Default: `''select(.reason == "compiler-artifact" and .executable != null and .profile.test == false)''` |
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -11,18 +11,6 @@
         "url": "https://github.com/ninegua/agent-rs/archive/4a22e590516bc79ec3c75a320f7941e7762ea098.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
-    "cargo": {
-        "branch": "nm/canonicalize-path-args",
-        "description": "The Rust package manager",
-        "homepage": "https://doc.rust-lang.org/cargo",
-        "owner": "nmattia",
-        "repo": "cargo",
-        "rev": "979904e092b39bb6aafe62a3ff3d994d3d8b6247",
-        "sha256": "078hhwii2qhdsm179xl249p0c2445yiyl5dv48kvx2372mc0g2gd",
-        "type": "tarball",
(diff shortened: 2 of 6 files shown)


Patryk27:
By the way:

> (that is, we have to scan the crates recursively, because bar might depend on crates/zar - and that's not an imaginary edge case, that's actually what happens inside Nushell, for instance.)

This is precisely why [my initial approach](https://github.com/nix-community/naersk/issues/274#issuecomment-1444406852) to the problem -  - doesn't actually work: nothing in the workspace's `Cargo.toml` lists `nu-glob`, it's only used transitively.

Patryk27:
cc @nmattia, @Anderssorby, @blitz - would you mind taking a look? 👀

nmattia:
hi @Patryk27, apologies for the delay! will have a look this week

README.md
```diff
@@ -213,7 +213,6 @@ process, rest is passed-through into `mkDerivation`.
 | `override` | An override for all derivations involved in the build. Default: `(x: x)` |
 | `overrideMain` | An override for the top-level (last, main) derivation. If both `override` and `overrideMain` are specified, _both_ will be applied to the top-level derivation. Default: `(x: x)` |
 | `singleStep` | When true, no intermediary (dependency-only) build is run. Enabling `singleStep` greatly reduces the incrementality of the builds. Default: `false` |
-| `targets` | The targets to build if the `Cargo.toml` is a virtual manifest. |
```
Patryk27, line 216: fwiw, this option wasn't actually used anyway - it was a no-op.

nix/sources.json
nmattia, line 1: Are those changes related to the new discovery algorithm?

Patryk27: Not directly - I was just adding sources for nushell (to create the test), realized that some of the items in `sources.json` are not actually used and thought it would be nice to tidy it up.

nmattia (approved):

nmattia:
I'd say let's merge and see if [who](https://xkcd.com/1172/) complains :)

edef1c:
This seems to ignore the `targets` option entirely. Cursory inspection with `targets = builtins.trace "targets evaluated" []` reveals that it is never even evaluated.

Patryk27:
Yes, that's correct:
https://github.com/nix-community/naersk/pull/276#discussion_r1139252946

Patryk27:
It seems like this `targets` attribute was never actually read - even if we go as far as cfcc584, this seems to be the only piece of code that accesses `targets`:
```nix
    # The members we want to build
    # (list of directory names)
    wantedMembers =
      lib.mapAttrsToList (member: _cargotoml: member) wantedMemberCargotomls;

    # Member path to cargotoml
    # (attrset from directory name to Nix object)
    wantedMemberCargotomls =
      let
        pred =
          if ! isWorkspace
          then (_member: _cargotoml: true)
          else
            if builtins.hasAttr "targets" attrs
            then (_member: cargotoml: lib.elem cargotoml.package.name attrs.targets)
            else (member: _cargotoml: member != ".");
      in
        lib.filterAttrs pred cargotomls;
```
... but they are not actually read anywhere.

@edef1c, did this pull request break your build or something? 👀

PR #281, opened 2023-03-22
Fix detection of symlinked Cargo.tomls

### LINKED ISSUE #280: New discovery algorithm doesn't handle symlinks
> I'd say let's merge and see if [who](https://xkcd.com/1172/) complains :)

In https://github.com/xfix/xfix.pw/blob/master/packages/pastebinrun.nix, my Rust program uses symlink for `Cargo.toml` which breaks with new algorithm that expects it to be a file.

Patryk27:
This commit fixes a bug introduced in #276 where symlinked Cargo.tomls wouldn't get detected due to readDir() reporting them as - well - _symlinks_ instead of regular files.

Note that we continue to keep the `if type == "directory"` condition for recursion which means that we'll fail to detect "nested" Cargo.tomls if someone decides to go wild and create workspace with symlinked crates.

That I'm not 100% sure on how to approach, but fortunately cases like those seem to be practically non-existent (and Crane apparently not supporting them builds a bit of confidence here as well).

Closes #280.


nmattia (approved):
LGTM, thanks for adding a reproduction too!

Issue #285, opened 2023-04-22
two step builds fail if CMake is used in a `build.rs`


wucke13:
Some crates with C bindings use CMake to actually build the C library in question. This breaks with two step builds, the error could look like this:

```
 CMake Error: The current CMakeCache.txt directory /build/source/target/x86_64-unknown-linux-gnu/release/build/oqs-sys-c5d43ceffffee31f/out/build/CMakeCache.txt is different than the directory /build/dummy-src/target/x86_64-unknown-linux-gnu/release/build/oqs-sys-c5d43ceffffee31f/out/build where CMakeCache.txt was created. This may result in binaries being created in the wrong place. If you are not sure, reedit the CMakeCache.txt
 ```
 
 Basically, CMake detects that the target folder was moved around, causing the build to fail. An easy hotfix would be:
 
 ```nix
 {
  #...
  overrideMain = x: x // {
    preBuild = x.preBuild + ''sed s_/dummy-src/_/source/_g --in-place "$(find -name CMakeCache.txt)"}'';
  };
}
```

Edit: Obviously its more elegant to use this as hotfix: `find -name CMakeCache.txt -exec sed s_/dummy-src/_/source/_g --in-place {} \;`

Hiding the evidence that trips of CMake.

Patryk27:
Hi, I've tried to replicate it (even with the same dependency, `oqs`), and everything seems to be working fine 👀 

Could you prepare some failing case? (preferably a repository I could `git clone` and `nix-build` / `nix build` locally to see)

wucke13:
@Patryk27 Yes, sure, above you find everything needed for a minimal example :smile:

Patryk27:
Huh, that's interesting - it _does_ fail when I build it with your expression, but this one seems to work:

```nix
{ ... }:
naersk.buildPackage {
  src = ./app;

  nativeBuildInputs = with pkgs; [ clang cmake ];
  buildInputs = with pkgs; [ openssl ];

  LIBCLANG_PATH = "${pkgs.llvmPackages.libclang.lib}/lib";
}
```

Could you check this on your side as well?

Patryk27:
Ah, right, that's probably because `CMakeCache.txt` dumps some envvars that are built differently when one uses `bindgenHook` vs writes `LIBCLANG_PATH` by hand; lemme see if we can safely bypass that on Naersk's side with your hook 🙂

PR #288, opened 2023-05-06
Support incremental builds with CMake

Patryk27:
When CMake builds a project, it generates a file called `CMakeCache.txt` that contains - among others - a dump of environmental variables.

This file is generated both for the deps-only- and the main- derivation, and if env-vars happen to differ between both¹, CMake will refuse to compile the latter derivation, saying:

```
CMake Error: The current CMakeCache.txt directory ... is different than the directory ... where CMakeCache.txt was created.
```

Solving this problem is easy - one just has to remove `CMakeCache.txt` to let CMake regenerate it.

Note that the original bug report (linked below) suggests to use `sed`, but that doesn't work on Darwin which uses a different convention for temporary paths - compare error messages for the failing derivations:

... on Linux:

```
CMake Error: The current CMakeCache.txt directory /build/source/target/release/build/oqs-sys-a19e222d9ef825d3/out/build/CMakeCache.txt is different than the directory /build/dummy-src/target/release/build/oqs-sys-a19e222d9ef825d3/out/build where CMakeCache.txt was created. This may result in binaries being created in the wrong place. If you are not sure, reedit the CMakeCache.txt
```

... and on Darwin:

```
CMake Error: The current CMakeCache.txt directory /tmp/nix-build-app-0.1.0.drv-5/source/target/release/build/oqs-sys-4b5e25b1671aa1e5/out/build/CMakeCache.txt is different than the directory /tmp/nix-build-app-deps-0.1.0.drv-0/source/target/release/build/oqs-sys-4b5e25b1671aa1e5/out/build where CMakeCache.txt was created. This may result in binaries being created in the wrong place. If you are not sure, reedit the CMakeCache.txt
```

As for the implementation, since two-step builds are a Naersk-specific construct, I think solving this transparently on Naersk's side instead of just adding a tip into README.md is the better approach here, since it keeps Naersk plug-and-play'able.

Closes https://github.com/nix-community/naersk/issues/285.

¹ which is not that difficult, e.g. that's how `pkgs.rustPlatform.bindgenHook` behaves

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -11,6 +11,18 @@
         "url": "https://github.com/ninegua/agent-rs/archive/4a22e590516bc79ec3c75a320f7941e7762ea098.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
+    "fenix": {
+        "branch": "main",
+        "description": "Rust toolchains and rust-analyzer nightly for Nix [maintainer=@figsoda]",
+        "homepage": "",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "06776a448cfdee8ed146305369c1810cd7c82799",
+        "sha256": "15z2mcrgzj9322xndk1xs29m23rdkfzbg2b0w0hva489rnq4mnc3",
+        "type": "tarball",
+        "url": "https://github.com/nix-community/fenix/archive/06776a448cfdee8ed146305369c1810cd7c82799.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
     "lorri": {
         "branch": "master",
         "description": "foo",
--- a/nix/sources.nix
+++ b/nix/sources.nix
@@ -10,29 +10,50 @@ let
     let
       name' = sanitizeName name + "-src";
     in
-      if spec.builtin or true then
-        builtins_fetchurl { inherit (spec) url sha256; name = name'; }
-      else
-        pkgs.fetchurl { inherit (spec) url sha256; name = name'; };
+    if spec.builtin or true then
+      builtins_fetchurl { inherit (spec) url sha256; name = name'; }
+    else
+      pkgs.fetchurl { inherit (spec) url sha256; name = name'; };
 
   fetch_tarball = pkgs: name: spec:
     let
       name' = sanitizeName name + "-src";
     in
-      if spec.builtin or true then
-        builtins_fetchTarball { name = name'; inherit (spec) url sha256; }
-      else
-        pkgs.fetchzip { name = name'; inherit (spec) url sha256; };
+    if spec.builtin or true then
+      builtins_fetchTarball { name = name'; inherit (spec) url sha256; }
+    else
+      pkgs.fetchzip { name = name'; inherit (spec) url sha256; };
 
   fetch_git = name: spec:
     let
       ref =
-        if spec ? ref then spec.ref else
+        spec.ref or (
           if spec ? branch then "refs/heads/${spec.branch}" else
-            if spec ? tag then "refs/tags/${spec.tag}" else
-              abort "In git source '${name}': Please specify `ref`, `tag` or `branch`!";
+          if spec ? tag then "refs/tags/${spec.tag}" else
+          abort "In git source '${name}': Please specify `ref`, `tag` or `branch`!"
+        );
+      submodules = spec.submodules or false;
+      submoduleArg =
+        let
+          nixSupportsSubmodules = builtins.compareVersions builtins.nixVersion "2.4" >= 0;
+          emptyArgWithWarning =
+            if submodules
+            then
+              builtins.trace
+                (
+                  "The niv input \"${name}\" uses submodules "
+                  + "but your nix's (${builtins.nixVersion}) builtins.fetchGit "
+                  + "does not support them"
+                )
+                { }
+            else { };
+        in
+        if nixSupportsSubmodules
+        then { inherit submodules; }
+        else emptyArgWithWarning;
     in
-      builtins.fetchGit { url = spec.repo; inherit (spec) rev; inherit ref; };
+    builtins.fetchGit
+      ({ url = spec.repo; inherit (spec) rev; inherit ref; } // submoduleArg);
 
   fetch_local = spec: spec.path;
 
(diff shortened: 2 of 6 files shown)


nix/sources.json
```diff
@@ -1,14 +1,26 @@
 {
     "agent-rs": {
```
Patryk27, line 2: `agent-rs` had to be bumped since we bumped `nixpkgs` which bumped `pkgs.openssl`, which was too newer for the older `agent-rs`'s `rust-openssl` to handle.

(i.e. compiling the older `agent-rs` using newer `nixpkgs` fails on `rust-openssl`)

nix/sources.json
```diff
@@ -1,14 +1,26 @@
...
         "repo": "agent-rs",
-        "rev": "4a22e590516bc79ec3c75a320f7941e7762ea098",
-        "sha256": "0sacddc34nlfgldqghlwchgzjki177h5dsgpmdv70cm8hfy0sg7l",
+        "rev": "71987142bb94a55ac0cb74a52396e01b6a142a16",
+        "sha256": "0hv3cg3f8jl0qpzm1qfrg1x7vbcqqv0mzjrgmh4g3w6f8xfj68is",
+        "type": "tarball",
+        "url": "https://github.com/dfinity/agent-rs/archive/71987142bb94a55ac0cb74a52396e01b6a142a16.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "fenix": {
```
Patryk27, line 14: I've switched `nixpkgs-mozilla` to `fenix`, since the latter works better on aarch64-darwin (e.g. it doesn't try to build rustc from scratch).

nix/sources.json
```diff
@@ -53,10 +65,10 @@
         "homepage": null,
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "6c2318e451fc0eaf338fb461d9bfcc99869de758",
-        "sha256": "1djwkvjnn26v0xw1swwh7dgmi0yl0b1kb8kansrdwk0jhhqbl7zq",
+        "rev": "89abc99d5b6351b2984c6126ef3bda7929905cb9",
```
Patryk27, line 68: I've bumped `nixpkgs`, since the older version doesn't contain `rustPlatform.bindgenHook` and I wanted to have a test as close to the original report as possible.

nix/sources.json
```diff
@@ -83,28 +95,16 @@
...
-    },
     "nushell": {
         "branch": "main",
         "description": "A new type of shell",
         "homepage": "https://www.nushell.sh/",
         "owner": "nushell",
         "repo": "nushell",
-        "rev": "85bfdca578157072e51e6972d370cfe63b0fda77",
-        "sha256": "0virlm3wpzsavdlci725iw0f8amr6ijxg9p2xp2q41zngjgb3w98",
+        "rev": "6dc7ff2335c9547182094451cd40c19e66c0678b",
```
Patryk27, line 104: I've bumped `nushell` since the older version had problems being built on aarch64-darwin.

nix/sources.nix
```diff
@@ -10,29 +10,50 @@ let
     let
```
Patryk27, line 10: Niv suggested to upgrade the file as well, so the diff here is just the result of `niv init`.

test/slow/agent-rs/default.nix
```diff
@@ -2,7 +2,6 @@
 
 naersk.buildPackage {
   src = sources.agent-rs;
-  doCheck = true;
```
Patryk27, line 5: `cargo test` had some problems on aarch64-darwin I didn't feel like investigating thoroughly at the moment 😇 

The build itself succeeds though.

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   linux:
     strategy:
         matrix:
-            nixpkgs: [ nixpkgs, nixpkgs-21.05, nixpkgs-21.11 ]
+            nixpkgs: [ nixpkgs ]
```
Patryk27, line 11: Multi-nixpkgs tests have become problematic because the older nixpkgs don't contain `rustPlatform.bindgenHook` and provide older OpenSSL - meaning that we can either use multi-nixpkgs only for fast-tests (like in here) _or_ have conditional tests (i.e. don't run the `oqs` test on non-recent nixpkgs).

Since reducing the matrix is the simpler approach (and we still make sure Naersk remains backwards-compatible through fast-tests), I think it's the better approach.

nmattia:
Missed the notification! I'll have a look ASAP

Patryk27:
Ping, @nmattia 🙂 

If you don't have time, it's alright, of course - I'm re-pinging just in case.

nmattia (commented):
@Patryk27 sorry, completely dropped the ball 😬 Lots going on privately

Is there any way you could split out the bits not directly related to CMake (formatting, package updates, etc)? That'd make it easier to review, right now I'm not 100% sure what's needed by the "CMake changes" and was is just clean up

Patryk27:
Okie, commit split 🙂

nmattia (approved):
Thanks a lot! Guess this had very little to do with CMake in the end :)

nmattia:
@Patryk27 sorry for the delay! Bit snowed under with personal stuff at the moment, but don't hesitate to ping me again if I forget :)

default.nix
```diff
@@ -62,13 +62,33 @@ let
...
+          # ```
+          # CMake Error: The current CMakeCache.txt directory ... is different than the directory ... where CMakeCache.txt was created.
+          # ```
+          drv = drv'.overrideAttrs (attrs: attrs // {
+            preBuild = (attrs.preBuild or "") + ''
+              find \
+                  -name CMakeCache.txt \
+                  -exec rm {} \;
+            '';
+          });
```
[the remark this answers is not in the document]
Patryk27: I've just checked it:
```nix
drv = drv'.overrideAttrs (attrs: attrs // {
  src = builtins.filterSource
    (name: type: (builtins.baseNameOf name) != "CMakeCache.txt")
    attrs.src;
});
```
... but it doesn't seem to work - most likely because the `CMakeCache.txt` comes not from `src` itself but rather from Naersk's `configurePhase`:

https://github.com/nix-community/naersk/blob/8507af04eb40c5520bd35d9ce6f9d2342cea5ad1/build.nix#L210

So perhaps a less confusing solution would be to put this `find ...` after that `for` loop, inside `configurePhase`? Something like this seems to work correctly:

```patch
diff --git a/build.nix b/build.nix
index cdfe73e..b324b9d 100644
--- a/build.nix
+++ b/build.nix
@@ -226,6 +226,11 @@ let
           fi
       done
 
+      find \
+          target \
+          -name CMakeCache.txt \
+          -exec rm {} \;
+
       mkdir -p $CARGO_HOME
 ```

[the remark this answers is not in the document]
nmattia: Actually, I think we could leave this up to the user, what do you think? And simply add an example/note of how to use naersk with CMake?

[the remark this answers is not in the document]
Patryk27: Sure, this sounds good, too - lemme adjust the merge request, then 🙂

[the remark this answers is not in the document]
Patryk27: alrighty, ready! :rocket:

Patryk27:
Ping, ping @nmattia - should be much easier to review now :smile:

I've had to bump Nix sources and add Fenix to get tests working on aarch64, but other than that, there's only the README.md change.

nmattia (approved):
Sorry for the delay, I was out for a bit! Let's gooo :)

PR #292, opened 2023-06-03
add option to run cargo check/test/clippy instead of building

icewind1991:
Motivation for running check/test/clippy through naersk instead of normally is that is allows using naersk's dependency caching to optimize the check/test/clippy run in CI as it already does for building.

With

```nix
packages.check = naersk'.buildPackage {
  src = ./.;
  mode = "check";
};
```

in your flake, and

```yaml
jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: cachix/install-nix-action@v20
      - uses: cachix/cachix-action@v12 # substitute with your favorite nix caching sollution
        with:
          name: mycache
          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
      - run: nix build .#check
```

in your github workflow. You get a nicely cached `cargo check` in your CI.


README.md
```diff
@@ -226,7 +226,9 @@ process, rest is passed-through into `mkDerivation`.
 | `copyTarget` | When true, the `target/` directory is copied to `$out`. Default: `false` |
 | `postInstall` | Optional hook to run after the compilation is done; inside this script, `$out/bin` contains compiled Rust binaries. Useful if your application needs e.g. custom environment variables, in which case you can simply run `wrapProgram $out/bin/your-app-name` in here. Default: `false` |
 | `usePureFromTOML` | Whether to use the `fromTOML` built-in or not. When set to `false` the python package `remarshal` is used instead (in a derivation) and the JSON output is read with `builtins.fromJSON`. This is a workaround for old versions of Nix. May be used safely from Nix 2.3 onwards where all bugs in `builtins.fromTOML` seem to have been fixed. Default: `true` |
-
+| `checkOnly` | When true, only runs `cargo check` during build instead of building the project, overwrites `cargoBuild`. Default: `false` |
+| `testOnly` | When true, only runs `cargo test` during build instead of building the project, overwrites `cargoBuild`. Default: `false` |
+| `testOnly` | When true, only runs `cargo test` during build instead of building the project, overwrites `cargoBuild`. Default: `false` |
```
Patryk27, line 231: This option seems to be duplicated here 🤔

build.nix
```diff
@@ -36,11 +36,15 @@
   #  Which drops the run-time dependency on the crates-io source thereby
   #  significantly reducing the Nix closure size.
 , removeReferencesToSrcFromDocs
+, checkOnly ? false
```
Patryk27, line 39: It looks like it makes sense to only have one of those active at once, so IMO we should either have an extra assertion:

```
if (checkOnly && (testOnly || clippyOnly)) ||
    (testOnly && (checkOnly || clippyOnly)) ||
    (clippyOnly && (checkOnly || testOnly))
  builtins.throw "Only one of `checkOnly`, `testOnly`, `clippyOnly` can be active."
```

... or switch to an enum-like approach:

```
# ...
, mode ? "build" # or `check`, or `test`, or `clippy`
# ...

if !builtins.elem mode ["build", "check", "test", "clippy"]
  builtins.throw "Unknown mode ${mode}, allowed modes: build, check, test, clippy"
```

Patryk27 (changes requested):
Looks neat, thanks for preparing it - I've left two comments to review 🙂

build.nix
icewind1991: Switch to a `mode` "enum".

Patryk27 (approved):
Lovely, thanks!

Issue #263, opened 2022-09-13
naersk rewrites `Cargo.toml` to something unparseable


GoldsteinE:
I tried to install `dylint` with these packages:

```nix
(naersk'.buildPackage {
  root = dylint;  # where `dylint` is a flake input pointing to `github:trailofbits/dylint/v2.0.12`
  cargoBuildOptions = x: x ++ [ "-p" "cargo-dylint" ];
  cargoTestOptions = x: x ++ [ "-p" "cargo-dylint" ];
})
(naersk'.buildPackage {
  root = dylint;
  cargoBuildOptions = x: x ++ [ "-p" "dylint-link" ];
  cargoTestOptions = x: x ++ [ "-p" "dylint-link" ];
})
```

I get this error:

```
error: failed to parse manifest at `/build/dummy-src/Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  TOML parse error at line 3, column 39
    |
  3 | metadata = { dylint = { libraries = [ {"path":"examples/*/*"} ] } }
    |                                       ^
  Unexpected `{`
  Expected `]`
```

Indeed, TOML in `dummy-src` looks like this:

```toml
[workspace]
members = [ "cargo-dylint", "dylint", "dylint-link", "internal", "utils/linting", "utils/testing" ]
metadata = { dylint = { libraries = [ {"path":"examples/*/*"} ] } }
```

While the actual TOML from the repo is correct:

```toml
# smoelius: `dylint_driver` is in its own workspace because it requires specific Rust components.

[workspace]
members = [
    "cargo-dylint",
    # "driver",
    "dylint",
    "dylint-link",
    "internal",
    "utils/linting",
    "utils/testing",
]

[workspace.metadata.dylint]
libraries = [
    { path = "examples/*/*" },
]
```

GoldsteinE:
It seems like this is the bug: it’s incorrect to just `toJSON` every val in list of attrs. The correct output should probably be

```
[[workspace.metadata.dylint.libraries]]
path = "examples/*/*"
```

Patryk27:
I see - thanks for investigating, I'll try fixing it 🙂

GoldsteinE:
As implementing full TOML serializer in Nix is complicated and further bugs may arise, maybe there is value in providing `usePureToTOML = false;` option to use external TOML serializer?

Xe:
I ran into this too:

```toml
[package.metadata.generate-rpm]
assets = [
[...]

PR #299, opened 2023-07-05
Use checksums for recreating crates.io's directory

Patryk27:
Closes https://github.com/nix-community/naersk/issues/138.

PR #300, opened 2023-07-14
Use TOML serializer from nixpkgs

Patryk27:
Requires https://github.com/nix-community/naersk/pull/288.
Closes https://github.com/nix-community/naersk/issues/263.


nmattia (dismissed):
Thanks! Didn't know this existed.

Patryk27:
I've rebased just now.

> Thanks! Didn't know this existed.

Yeah, the older version in nixpkgs was kinda bad (e.g. didn't support stuff like `[targets."cfg(...)"]`), but the current one looks pretty complete :smile:

Patryk27:
Ouch, something's failing on the CI - I'll take a look later 🤨

Patryk27:
Alright, the issue is that older nixpkgs' (< 23) `lib.formats.toml` doesn't properly handle cases like `[target."cfg(\"something\")"]`, which means that we can't rely on it there.

I've adjusted the logic to use our custom serializer (as it was) for older nixpkgs and rely on the nixpkgs' serializer only when nixpkgs are recent enough, with the intention of getting rid of our custom serializer when nixpkgs v23+ get more widely adopted.

Patryk27:
Huh, it's still failing, but the tests succeed on my Linux machine 🤨 -- lemme check on another one

Patryk27:
Alrighty, done!

nix/sources.json
```diff
@@ -60,15 +60,15 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "nixpkgs": {
-        "branch": "master",
+        "branch": "nixpkgs-unstable",
```
nmattia, line 63: any chance we can stay on `master`?

nmattia (dismissed):
Thanks a lot!

nix/sources.json
Patryk27: Sure, done 🙂

Patryk27:
hmm, analyzing failing tests 🔍

Patryk27:
Alright, got it!

I think https://github.com/nix-community/naersk/pull/299 caused us to (randomly) bump into https://github.com/rust-lang/cargo/pull/11188 where tl;dr it looks like when both crates.io & Git dependencies are packed together, as we do now:

https://github.com/nix-community/naersk/blob/d9a33d69a9c421d64c8d925428864e93be895dcc/build.nix#L91

... Cargo (sometimes, randomly) decides that Git-based dependencies must have checksums specified, which we don't do for them, since we don't know that checksum:

https://github.com/nix-community/naersk/blob/d9a33d69a9c421d64c8d925428864e93be895dcc/build.nix#L466

Creating two separate sources (one for crates.io and one for Git dependencies), as I've done in the recent commit just now, seems to avoid the issue (I've checked on aarch64-darwin, x86_64-linux NixOS + x86_64-linux Ubuntu, and everything passed on the first time now).

nmattia (approved):
Thanks for all the work!

Issue #305, opened 2023-08-16
dummy-src build failed with "cp: missing file operand"


cole-h:
```
dummy-src> cp: missing file operand
dummy-src> Try 'cp --help' for more information.
note: keeping build directory '/tmp/nix-build-dummy-src.drv-4'
error: builder for '/nix/store/y62b65nl4bq3zwj6kqm5a7387d2b9843-dummy-src.drv' failed with exit code 1;
       last 2 log lines:
       > cp: missing file operand
       > Try 'cp --help' for more information.
       For full logs, run 'nix log /nix/store/y62b65nl4bq3zwj6kqm5a7387d2b9843-dummy-src.drv'.
```

This happened when I updated naersk from d9a33d69a9c421d64c8d925428864e93be895dcc to 275010712ce41dff66634f9910bc1f085239b370.

Patryk27:
Whoopsie, would you mind trying on 9bbe32cccfff5a44b7b75fb818e7739d500a07a7?

cole-h:
Same issue, I'm afraid.

You should be able to reproduce with `nix build github:cole-h/nixos-config/naersk-debug#debug.x86_64-linux`.

Patryk27:
Okie, confirmed, it fails on my machine as well - I'll take a look tomorrow.

cole-h:
Just chiming in that this is confirmed fixed now, thanks :)

(For posterity: I've deleted the `naersk-debug` branch on my repo, so the above `nix build` won't work)

PR #306, opened 2023-08-17
Fix support for `.cargo/config`

Patryk27:
Before https://github.com/nix-community/naersk/pull/300, our `builtinz.toTOML` used to return a string representing the serialized document which was later redirected into a file:

https://github.com/nix-community/naersk/pull/300/files#diff-b6b537316f2d29c8faf178a110366796811d1bc72f694262c7d2efad79aa984bL238

Over that merge request, this was changed to return a path, to make it consistent with how nixpkgs' formatters work - but I haven't noticed one place that still _read_ the file instead of returning a path (which was fine before, since the code did `echo ... > ...`, but is not a correct thing to do for `cp`).

Closes https://github.com/nix-community/naersk/issues/305.


nmattia (dismissed):
Thanks for the fix, and the test!

Patryk27:
Small fix: tests for nixpkgs-21.05 used too old Cargo, from before support for custom environmental variables was introduced, so I've had to adjust the new test a bit.

Patryk27:
(cc @nmattia, since GitHub requires another ✅ here :smile:)

nmattia (approved):
There you go :)

PR #315, opened 2023-10-24
readme: Add maintainer status

Patryk27:
I don't have much time nowadays to manage Naersk, so it's time to give the scepter :-)


nmattia (approved):
@Patryk27 thank you for all the work you put in over the past, what was it, two-three years already! ❤️

PR #317, opened 2023-10-28
chore: formatted code and fixed lints

AxelSilverdew:
I took the liberty of formatting all the code with alejandra and resolving a few lint warnings that were thrown by statix.

This should make it a wee bit easier to contribute code to the repo.


Patryk27:
While I'm not necessarily the number one fan of this particular formatting style (more of a nixpkgs-fmt guy myself), I think as long as we have something consistent, it's nice -- would it be possible to attach some formatting check into the pipeline as well, so that misformatted code can't be committed?

cc @nmattia, ofc.

AxelSilverdew:
Oh yeah, it should be pretty easy to add a format and lint check to the pipeline.
I'll push a commit adding that in a little bit.

nmattia:
Oh, I had never heard of Alejandra! I don't have a strong opinion for one style or another, but since this is technically a nix-community project, might be less confusing to use nixpkgs-fmt? or is alejandra the accepted default formatter now?

> would it be possible to attach some formatting check into the pipeline 

would even go further and say: let's not format anything unless there are CI checks :)

AxelSilverdew:
Also for the record, I don't have a preference between nixpkgs-fmt or alejandra, so I'm fine with whatever as long as the outputs are consistent. :D

Issue #224, opened 2022-01-12
`nix run .#` in workspace doesn't work unless `version =` is set


dpc:
Without `version = ` the binary will be `/nix/store/0xdfyy5fsl4pisiqki0slc2ajr62jdfg-<name>-unknown/bin/<name>` but `nix run` will try to execute ` `/nix/store/0xdfyy5fsl4pisiqki0slc2ajr62jdfg-<name>-unknown/bin/<name>-unknown`

dpc:
Code examples (the fix: https://github.com/crev-dev/cargo-crev/commit/3d248466994b5f370708834cc52f70505eabdf4c)

blitz:
This is interesting. It seems Nix automatically strips version numbers, but fails (understandably) for the "unknown" suffix. We can either:

- drop "-unknown" in the derivation names,
- set `meta.programName` to `${pname}`.

Sean1708:
Hmmm.... I'm running into a similar problem where instead of trying to run `<name>-unknown` it's trying to execute `<name>-<version>` but the binary is just called `<name>`.  Setting version doesn't fix it, instead I have to strip the version manually in the `exePath` argument to `mkApp`.

Sean1708:
In case a little more info is useful, my _working_ `flake.nix` is

```nix
{

  inputs = {
    naersk.url = "github:nmattia/naersk/master";
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    utils.url = "github:numtide/flake-utils";
  };

  outputs = { self, nixpkgs, utils, naersk }:
    utils.lib.eachDefaultSystem (system:
      let
        pkgs = import nixpkgs { inherit system; };
        naersk-lib = pkgs.callPackage naersk { };
        nativeBuildInputs = with pkgs; [ gtk4 pkg-config ];
      in {
        defaultPackage = naersk-lib.buildPackage {
          root = ./.;

          inherit nativeBuildInputs;
        };

        defaultApp = let
          drv = self.defaultPackage."${system}";
          name = pkgs.lib.strings.removeSuffix ("-" + drv.version) drv.name;
        in utils.lib.mkApp {
          inherit drv;
          # TODO: https://github.com/nix-community/naersk/issues/224
          exePath = "/bin/${name}";
        };

        devShell = with pkgs;
          mkShell {
            buildInputs =
              [ cargo rust-analyzer rustc rustfmt rustPackages.clippy ];

[...]

PR #322, opened 2024-04-10
Fixes nix-community/naersk#224

noebm:
`name` will still be defined by mkDerivation when `pname` and `version` are defined and this is actually preferred according to RFC0035.

By defining `pname`, `flake-utils.lib.mkApp` will prefer this over `name` when calling the executable.

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -76,8 +76,8 @@ let
       { inherit lib writeText remarshal runCommandLocal formats; };
 
   drvAttrs = {
-    name = "${pname}-${version}";
     inherit
+      pname
       src
       version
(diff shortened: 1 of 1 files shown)


nmattia (approved):
Thank you @noebm ! 

Are you still using naersk? I was actually about to archive the repo, see discussion here: https://github.com/nix-community/naersk/pull/214

noebm:
Oh I just started using it. So you would suggest switch to crane?

nmattia:
> So you would suggest switch to crane?

Definitely! Naersk is very much unmaintained at this point!

PR #325, opened 2024-04-16
Fix .cargo-home/config warning on nightly, bump probe-rs

exFalso:
This PR fixes the nightly warning
```
warning: `/build/probe-rs-src/.cargo-home/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
```

Also bumps the probe-rs sources and the rust nightly version used for testing it

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -244,6 +244,7 @@ let
       mkdir -p $CARGO_HOME
 
       cp "$cargoconfig" $CARGO_HOME/config.toml
+      # symlink for backwards compatibility with older cargo
       ln -s ./config.toml $CARGO_HOME/config
 
       runHook postConfigure
--- a/build.nix
+++ b/build.nix
@@ -243,7 +243,8 @@ let
       export CARGO_HOME=''${CARGO_HOME:-$PWD/.cargo-home}
       mkdir -p $CARGO_HOME
 
-      cp "$cargoconfig" $CARGO_HOME/config
+      cp "$cargoconfig" $CARGO_HOME/config.toml
+      ln -s ./config.toml $CARGO_HOME/config
 
       runHook postConfigure
     '';
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -137,10 +137,10 @@
         "homepage": "https://probe.rs",
         "owner": "probe-rs",
         "repo": "probe-rs",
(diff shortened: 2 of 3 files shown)


nmattia:
Thank you @exFalso !

Out of curiosity, are you actively using naersk? I'm planning on marking it as deprecated: https://github.com/nix-community/naersk/pull/214#issuecomment-1963774309

exFalso:
Oh hmm hmm, yes we've been using it for quite some time now(2020 i think?).

At the time we picked it, it was the only nix-cargo project that worked *at all* with our setup. We use nightly rust with special targets (SGX and WASM), we have workspaces, local dependencies, forks of dependencies, git dependencies etc. And we wanted plain cargo to "just work" during development. None of the other tools worked, although `crane` did not exist at the time.

`naersk` just sort of works in its current state (I guess besides these smaller oddities as rust evolves over time), so we probably won't spend dev resources on a switch any time soon. If you do deprecate it, then thank you for the project and the maintenance effort so far! It's been (and still is) one of the most important workhorses for our builds.

build.nix
```diff
@@ -243,7 +243,8 @@ let
       export CARGO_HOME=''${CARGO_HOME:-$PWD/.cargo-home}
       mkdir -p $CARGO_HOME
 
-      cp "$cargoconfig" $CARGO_HOME/config
+      cp "$cargoconfig" $CARGO_HOME/config.toml
+      ln -s ./config.toml $CARGO_HOME/config
```
nmattia, line 247:
```suggestion
      # symlink for backwards compatibility with older cargo
      ln -s ./config.toml $CARGO_HOME/config
```

nmattia (dismissed):
LGTM! Minor suggestion

nmattia:
> It's been (and still is) one of the most important workhorses for our builds.

That's very nice to hear!

Then I might keep it around for a bit longer. Might be a good idea to give e.g. crane a try as it might prove more future proof though. I'm actually surprised to hear that naersk can somewhat keep up with recent rust changes; I think it says a lot about the backwards compatibility of the rust platform!

nmattia (approved):

PR #326, opened 2024-05-16
Crate specific overrides (eg. building OpenSSL without errors out of the box)

### LINKED ISSUE #6: Have some extra packages per crate
Some crates, like openssl-sys, will always require `openssl` and `pkgconfig`. They could be provided by default.

FireFragment:
This PR adds a new system of _crate-specific overrides_, which allows to automatically adjust build behavior to make some crates working. \
This will probably be mainly adding non-rust system dependencies of crates automatically as `buildInputs`, so they work out-of-the-box for users.

In this PR, there is just one such override (if `openssl-sys` is in dependencies, automatically provide `pkg-config` in `nativeBuildInputs` and `openssl` in `buildInputs`), but more can be now added easily in future by editing `crate_specific.nix`. 

  - Fixes #6 

### Summary of the changes

 - Added the file `crate_specific.nix` for defining the crate-specific overrides easily
 - Added `autoCrateSpecificOverrides` option for enabling/disabling this feature
 - Added a crate with a dependency on OpenSSL to test, whether it works


crate_specific.nix
```diff
@@ -0,0 +1,16 @@
+# This file contains crate-specific build inputs.
+#
+# Each of them is an attribute with name of the crate and value being a function.
+# This function gets as arguments:
+#  - `crateInfo` with information about the crate, such as sha256 or version
+#
+# Currently supported fields to return are: `buildInputs`, `nativeBuildInputs`
```
nmattia, line 7: And the returned `buildInputs` are simply added to the other build inputs, right?

README.md
```diff
@@ -267,6 +267,7 @@ Note that you shouldn't call `overrideAttrs` on a derivation built by Naersk
 | `postInstall` | Optional hook to run after the compilation is done; inside this script, `$out/bin` contains compiled Rust binaries. Useful if your application needs e.g. custom environment variables, in which case you can simply run `wrapProgram $out/bin/your-app-name` in here. Default: `false` |
 | `usePureFromTOML` | Whether to use the `fromTOML` built-in or not. When set to `false` the python package `remarshal` is used instead (in a derivation) and the JSON output is read with `builtins.fromJSON`. This is a workaround for old versions of Nix. May be used safely from Nix 2.3 onwards where all bugs in `builtins.fromTOML` seem to have been fixed. Default: `true` |
 | `mode` | What to do when building the derivation. Either `build`, `check`, `test`, `fmt` or `clippy`. <br/> When set to something other than `build`, no binaries are generated. Default: `"build"` |
+| `autoCrateSpecificOverrides` | Whether to automatically apply crate-specific overrides, mainly additional `buildInputs` for dependencies. <br /> For example, if you use the `openssl` crate, `pkgs.pkg-config` and `pkgs.openssl` are automatically added as buildInputs. Default: `true` |
```
nmattia, line 270: I think having it enabled by default might create issues for users that already build with openssl and already add dependencies manually. WDYT?

nmattia (commented):
Very cool!

crate_specific.nix
FireFragment: Yes, here:
https://github.com/nix-community/naersk/pull/326/files#diff-b6b537316f2d29c8faf178a110366796811d1bc72f694262c7d2efad79aa984bR146

README.md
FireFragment: One idea I have is to emit a warning if there is already a derivation in `buildInputs` with the same name. It would not work in 100% of the cases, but it should work most of the time. Or we can just disable it by default.

nmattia: I think having it enabled by default is much nicer!

nmattia:
@FireFragment looks like there's a cargo issue in the test, do they run successfully for you?

```
       >   failed to update replaced source registry `[https://github.com/rust-lang/crates.io-index`](https://github.com/rust-lang/crates.io-index%60)
       >
       > Caused by:
       >   failed to parse manifest at `/nix/store/shh4214g8hn6jakwa8vrhbmqpx8fzj67-crates-io-dependencies/bitflags-2.5.0-cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1/Cargo.toml`
       >
       > Caused by:
       >   feature `edition2021` is required
       >
       >   consider adding `cargo-features = ["edition2021"]` to the manifest
```

https://github.com/nix-community/naersk/actions/runs/9117568700/job/25091230219?pr=326#step:4:3064

FireFragment:
It looks like that in nixpkgs 22 and lower, the cargo and rust version is too old to compile the `openssl` crate, so I just disabled the test on nixpkgs versions lower than 23 [here](https://github.com/nix-community/naersk/pull/326/commits/909f97aa958b6df4ed14d746a076290c047e6c0f).

Now it passes: https://github.com/FireFragment/naersk/actions/runs/9275182581

nmattia (approved):
Thanks a ton for doing this @FireFragment, much appreciated!

Issue #146, opened 2021-02-08
Support -Z build-std


Stupremee:
See
- https://github.com/rust-lang/wg-cargo-std-aware
- https://doc.rust-lang.org/cargo/reference/unstable.html#build-std
- https://internals.rust-lang.org/t/dogfooding-z-build-std-in-rustbuild/13775/13

Currently, if you try to build a project that uses `build-std`, you'll get errors about dependencies that are not found, but definitely
exist. 

### Example project

<details><summary>.cargo/config.toml</summary>
<p>

```toml
[build]
target = "riscv64gc-unknown-none-elf"

[unstable]
build-std = ["core"]
```
</p>
</details>

<details><summary>src/lib.rs</summary>
<p>

```rust
#![no_std]
```
</p>
</details>

<details><summary>default.nix</summary>
<p>

```nix
let
  sources = import ./nix/sources.nix;
  nixpkgs-mozilla = import sources.nixpkgs-mozilla;
  rust = pkgs.latest.rustChannels.nightly.rust.override {
    targets = [ "riscv64gc-unknown-none-elf" ];
    extensions = [ "rust-src" ];
  };
  pkgs = import sources.nixpkgs {
    overlays = [
      nixpkgs-mozilla
      (self: super: {
        rustc = rust;
        cargo = rust;
      })
    ];
  };
  naersk = pkgs.callPackage sources.naersk { };
in naersk.buildPackage ./.
```
</p>
</details>

If I try to build this project, I get the following error:

<details><summary>Error</summary>
<p>

```
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking sources
unpacking source archive /nix/store/8aaja1qq93a0bvv48bij1n5gcwsblidy-foo
source root is foo
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
[naersk] cargo_version (read): 1.51.0-nightly (34170fcd6 2021-02-04)
[naersk] cargo_message_format (set): json-diagnostic-rendered-ansi
[naersk] cargo_release: --release
[naersk] cargo_options: -Z unstable-options
[naersk] cargo_build_options: $cargo_release -j "$NIX_BUILD_CORES" --out-dir out --message-format=$cargo_message_format
[naersk] cargo_test_options: $cargo_release -j "$NIX_BUILD_CORES"
[naersk] RUST_TEST_THREADS: 16
[...]

PR #328, opened 2024-06-18
Add additionalCargoLock argument

bouk:
This allows for an extra Cargo.lock, which is useful when using build-std to ensure the requires crates are available.

Related to: #146

This can be used like:

```
naersk.buildPackage {
  ...
  additionalCargoLock = ./rust-std-Cargo.lock;
}
```

Which will then ensure all the sources that are required to build rust-std are included. The Cargo.lock can be found at https://github.com/rust-lang/rust/blob/master/Cargo.lock


nmattia (approved):
Thank you!

PR #335, opened 2024-12-04
Drop macOS SDKs that follow "the old way"

grahamc:
Nixpkgs now has much more reasonable handling of macOS SDKs, and this old default is not generally needed  for builds anymore.

I think having these SDKs in the inputs is a bit of a nuisance now, since it means there are mixed framework versions in the closure.

See: https://discourse.nixos.org/t/on-the-future-of-darwin-sdks-or-how-you-can-stop-worrying-and-put-the-sdk-in-build-inputs/50574
See: https://github.com/NixOS/nixpkgs/pull/346043

Note: this is also tagged at https://github.com/DeterminateSystems/naersk/releases/tag/apple-sdks for use.

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -58,7 +58,6 @@
 , lib
 , rsync
 , jq
-, darwin
 , writeText
 , runCommandLocal
 , remarshal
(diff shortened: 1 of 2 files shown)


nmattia:
Oh that's very good news! I haven't been following changes in nixpkgs but this definitely cleans things up a fair bit.

Thanks a lot!

nmattia (approved):

PR #350, opened 2025-04-04
don't use indirect nixpkgs references

Mic92:
Those are deprecated in Nix:

https://github.com/NixOS/nix/pull/12068


nmattia (approved):
Thanks!

Issue #353, opened 2025-04-22
Failing cargo test causes jq parse error


cpick:
When I use `nix build` to have naersk run `cargo test` (via the `mode = "test"` attribute) it will succeed as long as all tests pass, but if a test fails then naersk prints a `jq: parse error` instead of the test's output:
```sh
$ nix build
error: builder for '/nix/store/b4v9mxfw389jm9syglzl0yd0jww5hrjs-naersk-fail-test-parse-0.1.0.drv' failed with exit code 5;
       last 25 log lines:
       > unpacking source archive /nix/store/ajhnk2r5iaxqdpdz62iwpdh41s7pn54v-source
       > source root is source
       > Running phase: patchPhase
       > Running phase: updateAutotoolsGnuConfigScriptsPhase
       > Running phase: configurePhase
       > [naersk] cargo_version (read): 1.86.0 (adf9b6ad1 2025-02-28)
       > [naersk] cargo_message_format (set): json-diagnostic-rendered-ansi
       > [naersk] cargo_release: --release
       > [naersk] cargo_options:
       > [naersk] cargo_build_options: $cargo_release -j "$NIX_BUILD_CORES" --message-format=$cargo_message_format
       > [naersk] cargo_test_options: $cargo_release -j "$NIX_BUILD_CORES"
       > [naersk] RUST_TEST_THREADS: 8
       > [naersk] cargo_bins_jq_filter: select(.reason == "compiler-artifact" and .executable != null and .profile.test == false)
       > [naersk] cargo_build_output_json (created): /build/tmp.KmtJ34tdMN
       > [naersk] RUSTFLAGS:
       > [naersk] CARGO_BUILD_RUSTFLAGS:
       > [naersk] CARGO_BUILD_RUSTFLAGS (updated):  --remap-path-prefix /nix/store/hyn6f6fr11xbjbxcp13jw7hzx8s4w63n-crates-io-dependencies=/sources --remap-path-prefix /nix/store/gri5skg1j3wslkyakxmb9rfcn8znq6n8-git-dependencies=/sources
       > [naersk] pre-installing dep /nix/store/jbkqa93in1j77pkc1836q6wf5lca9g4d-naersk-fail-test-parse-deps-0.1.0
       > Running phase: buildPhase
       > cargo test $cargo_release -j "$NIX_BUILD_CORES"
       >    Compiling naersk-fail-test-parse v0.1.0 (/build/source)
       >     Finished `release` profile [optimized] target(s) in 0.08s
[...]

PR #354, opened 2025-04-22
Fix jq parse error when cargo test runs a failing test

cpick:
Relay non-JSON lines in cargo output

Forward both JSON log messages from `cargo test` and non-JSON messagse from the test binary itself.

(Note: the test binaries do have a `--format` option, but the "json" format is unstable and so can only be used by nightly `cargo`)

Fixes: https://github.com/nix-community/naersk/issues/353

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -258,6 +258,8 @@ let
       logRun ${cargoCommand} || cargo_ec="$?"
 
       if [ "$cargo_ec" -ne "0" ]; then
+      # we allow non-json (with jq -R, try, catch) as tests might also produce output
+      # https://github.com/nix-community/naersk/issues/353
         cat "$cargo_build_output_json" | jq -cMrR '. as $line | try (fromjson | select(.message.rendered != null) | .message.rendered) catch $line'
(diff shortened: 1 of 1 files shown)


build.nix
```diff
@@ -258,7 +258,7 @@ let
       logRun ${cargoCommand} || cargo_ec="$?"
 
       if [ "$cargo_ec" -ne "0" ]; then
```
nmattia, line 260:
```suggestion
      if [ "$cargo_ec" -ne "0" ]; then
      # we allow non-json (with jq -R, try, catch) as tests might also produce output
      # https://github.com/nix-community/naersk/issues/353
```

nmattia (approved):
Hey sorry for the slow feedback, I was on holiday.

> --raw-input / -R:
> Don't parse the input as JSON. Instead, each line of text is passed to the filter as a string. If combined with --slurp, then the entire input is passed to the filter as a single long string.

Well TIL! Thanks!

nmattia (approved):

nmattia:
Thanks!

Issue #371, opened 2025-11-16
Windows cross-compilation example does not work with the latest nixpkgs


link2xt:
To reproduce the problem, clone this repository, then go to `naersk/examples/cross-windows`, run `nix flake update` and then `nix build`.

Bulid fails:
```
$ nix build
warning: Git tree '/home/link2xt/src/naersk' is dirty
error:
       … while calling the 'derivationStrict' builtin

         at /derivation-internal.nix:9:12:

            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'cross-windows-0.1.0'
         whose name attribute is located at /nix/store/9g7vwfbsk7hcp12vrm8a82cnv6y9v8fb-source/pkgs/stdenv/generic/make-derivation.nix:468:13

       … while evaluating attribute 'builtDependencies' of derivation 'cross-windows-0.1.0'

         at /nix/store/hvi5hs0v1msi8sfsndxf7xh6zn84dg8r-source/build.nix:142:12:

          141|
          142|     inherit builtDependencies;
             |            ^
          143|

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: Package ‘mingw_w64-pthreads-13.0.0’ in /nix/store/k5s8fqk94zzga84iym6wvaqnhmq7ykwf-source/pkgs/os-specific/windows/mingw-w64/headers.nix:35 is not available on the requested hostPlatform:
         hostPlatform.system = "x86_64-linux"
         package.meta.platforms = [
           "aarch64-windows"
           "x86_64-windows"
           "i686-windows"
         ]
         package.meta.badPlatforms = [ ]
       , refusing to evaluate.

       a) To temporarily allow packages that are unsupported for this system, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1

          Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
                then pass `--impure` in order to allow use of environment variables.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowUnsupportedSystem = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowUnsupportedSystem = true; }
       to ~/.config/nixpkgs/config.nix.
```

link2xt:
[...]

PR #373, opened 2025-11-17
Use only one version of nixpkgs in cross-windows example

link2xt:
Example still builds. This does not update nixpkgs, but should make debugging https://github.com/nix-community/naersk/issues/371 easier.


nmattia (approved):
Thanks! Don't have a windows machine to try it on but I'm guessing you did build it.

link2xt:
@nmattia I only tried it on Linux, it builds on Linux.

PR #388, opened 2026-05-21
fix: remove `package.metabuild` when building dummy src

### LINKED ISSUE #387: `metabuild` breaks dependency build
Building a crate that uses [the metabuild feature](https://doc.rust-lang.org/1.94.0/cargo/reference/unstable.html#metabuild) (unstable) fails unless `singleStep` is set:

```
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
cargo build $cargo_release -j "$NIX_BUILD_CORES" --message-format=$cargo_message_format -p my_crate3
warning: `/build/dummy-src/.cargo/config` is deprecated in favor of `config.toml`
  |
  = help: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
error: failed to load manifest for workspace member `/build/dummy-src/crates/my_crate2`
referenced via `crates/*` by workspace at `/build/dummy-src/Cargo.toml`

Caused by:
  failed to load manifest for dependency `my_crate`

Caused by:
  failed to parse manifest at `/build/dummy-src/crates/my_crate/Cargo.toml`

Caused by:
  cannot specify both `metabuild` and `build`
[naersk] cargo returned with exit code 101, exiting
```

kawadakk:
Fixes #387.

Cargo's [metabuild feature][1] (unstable) defines declarative build scripts. This conflicts with normal build scripts that we generate for intermediary (dependency-only) builds, so we need to remove the `package.metabuild` key to avoid build failures.

[1]: https://doc.rust-lang.org/1.94.0/cargo/reference/unstable.html#metabuild


nmattia (approved):
Thanks a lot! I was able to reproduce the issue without fenix.

PR #391, opened 2026-05-27
Use static.crates.io to download crates

### LINKED ISSUE #390: crates.io returns 403 when downloading dependencies (2026-05-27)
Building with `naersk` now results in errors like this:
```
download-netlink-packet-utils>   0      0   0      0   0      0      0      0                              0
download-netlink-packet-utils> curl: (22) The requested URL returned error: 403
download-netlink-packet-utils> Warning: Problem (retrying all errors). Retrying in 4 seconds. 1 retry left.
download-ed25519-dalek>   0      0   0      0   0      0      0      0                              0
download-ed25519-dalek> curl: (22) The requested URL returned error: 403
download-ed25519-dalek> Warning: Problem (retrying all errors). Retrying in 4 seconds. 1 retry left.
download-gif>   0      0   0      0   0      0      0      0                              0
download-gif> curl: (22) The requested URL returned error: 403
download-gif> Warning: Problem (retrying all errors). Retrying in 4 seconds. 1 retry left.
```

Here `fetchurl` is used in `naersk`:
https://github.com/nix-community/naersk/blob/efdddff9ff4d8e7d0056d57ec67dac50f75ab8f6/build.nix#L435-L440

There are attempts to fix the problem in nixpkgs by setting the user agent, see https://github.com/NixOS/nixpkgs/pull/524986 and https://github.com/NixOS/nixpkgs/pull/512735 and https://github.com/rust-lang/crates.io/issues/13482 and other relate issues.

link2xt:
Closes #390 

This is similar to https://github.com/NixOS/nixpkgs/pull/524985

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -435,7 +435,7 @@ let
       crate = fetchurl {
         inherit sha256;
 
-        url = "${cratesDownloadUrl}/api/v1/crates/${name}/${version}/download";
+        url = "${cratesDownloadUrl}/${name}/${version}/download";
         name = "download-${name}-${version}";
       };
 
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ Note that you shouldn't call `overrideAttrs` on a derivation built by Naersk
 | `gitAllRefs` | Whether to fetch all refs while fetching Git dependencies. Useful if the wanted revision isn't in the default branch. Requires Nix 2.4+. Default: `false` |
 | `gitSubmodules` | Whether to fetch submodules while fetching Git dependencies. Requires Nix 2.4+. Default: `false` |
 | `additionalCargoLock` | Additional cargo lock used to specify crates required for build |
-| `cratesDownloadUrl` | Url for downloading crates from an alternative source Default: `"https://crates.io"` |
+| `cratesDownloadUrl` | Url for downloading crates from an alternative source Default: `""https://static.crates.io/crates"` |
 | `cargoBuild` | The command to use for the build. The argument must be a function modifying the default value. <br/> Default: `''cargo $cargo_options build $cargo_build_options >> $cargo_build_output_json''` |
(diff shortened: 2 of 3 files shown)


build.nix
```diff
@@ -435,7 +435,7 @@ let
       crate = fetchurl {
         inherit sha256;
 
-        url = "${cratesDownloadUrl}/api/v1/crates/${name}/${version}/download";
```
link2xt, line 438: This is a breaking change for users of `cratesDownloadUrl` because we don't append `api/v1/crates` anymore.

link2xt: But I guess users will figure it out and can add this to the URL manually.

lriesebos:
I was keeping an eye open for this PR since I was suffering from this issue, but that seems to have solved itself automatically for me at least? maybe crates.io reverted some change that caused this problem in the first place?

komar007:
@lriesebos 
I am pretty sure I had the same automatic fix happen to me.
Here is my failed build: https://github.com/komar007/komar-os/actions/runs/26474920347/job/78339869321
It failed on anstream-0.6.19 which is in the Cargo.lock of the current nix-community/comma: https://github.com/nix-community/comma/blob/218175027871fc8f51735d3e5ab38f945e3a0e2f/Cargo.lock#L16 and I am rather certain I don't have any instance of that version used elsewhere. Comma uses naersk to build and at some point setting comma's naersk input to this PR fixed the issue: https://github.com/komar007/komar-os/commit/59d9f2b9415ef900e3e0c230571f425b296cc2fc.
Now, my HM config happily builds without it.
Unfortunately, a re-run of the original failed build fails for a different reason (another crate, then built with nixpkgs-25.11 when it was still affected), so I cannot directly prove the change in crates.io API, because I would be building a different derivation.

nmattia (approved):
LGTM, thanks a lot for the fix!

PR #393, opened 2026-06-08
Fix .cargo-home/config and config.toml warning

dly04:
This PR fixes the warning
```
warning: Both `/build/source/.cargo-home/config` and `/build/source/.cargo-home/config.toml` exist. Using `/build/source/.cargo-home/config`
```
Although it hinted us to create a symlink in #325 , it seems that now it would prefer only one of `.cargo-home/config` and `.cargo-home.toml` .

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -244,8 +244,6 @@ let
       mkdir -p $CARGO_HOME
 
       cp "$cargoconfig" $CARGO_HOME/config.toml
-      # symlink for backwards compatibility with older cargo
-      ln -s ./config.toml $CARGO_HOME/config
 
       runHook postConfigure
     '';


build.nix
```diff
@@ -243,9 +243,15 @@ let
       export CARGO_HOME=''${CARGO_HOME:-$PWD/.cargo-home}
       mkdir -p $CARGO_HOME
 
-      cp "$cargoconfig" $CARGO_HOME/config.toml
-      # symlink for backwards compatibility with older cargo
-      ln -s ./config.toml $CARGO_HOME/config
+      # .cargo-home/config.toml is introduced and preferred in 1.39: 
+      # https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html#cargo-139-2019-11-07
+      IFS=. read -r cargo_version_major cargo_version_minor _ <<< "$cargo_version"
+      if (( cargo_version_major > 1 || cargo_version_major == 1 && cargo_version_minor >= 39 ));
```
nmattia: Thanks! I'd say let's just use `config.toml`. It's been supported for a while; we can re-add the `if` if anyone complains.

dly04: Yes sure.

nmattia (approved):

nmattia:
Thank you @dly04 !

